PR #67, opened 2022-07-26
NTP-378: Cost per subject content

benagain:

* Remove the entire column for a tuition type if it is not provided.
* Trim prices where they are whole pounds.

Before: 
image

After: 
image

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-378

## Things to check

### DIFF
--- a/UI/Pages/TuitionPartner.cshtml
+++ b/UI/Pages/TuitionPartner.cshtml
@@ -132,8 +132,8 @@
 		    <thead class="govuk-table__head">
 		    <tr class="govuk-table__row">
 			    <th scope="col" class="govuk-table__header">Tuition group size</th>
-			    <th scope="col" class="govuk-table__header">In school</th>
-			    <th scope="col" class="govuk-table__header">Online</th>
+			    <th show-if="@Model.Data.Prices.ContainsInSchoolPrice()" scope="col" class="govuk-table__header">In school</th>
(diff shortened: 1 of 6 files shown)


UI/Pages/TuitionPartner.cshtml
```diff
@@ -132,31 +132,22 @@
 		    <thead class="govuk-table__head">
```
iterative-it, line 132: I believe there's some additional content that goes above the table according to the ticket:

```
<p>The costs shown are per pupil per hour. 
[Next line] All subjects are the same price but depend on tuition group size.
```

However the second line talks about subjects having the same price which wouldn't work for the ones that do vary by subject. Can you go back to the BAs/content and ask what should be displayed when the price does vary by subject?

iterative-it (changes requested):

UI/Pages/TuitionPartner.cshtml
benagain: Yep, I missed that.  I have added the extra content.  

I couldn't find any examples of current tuition partners that varied the price by subject, but I will ask the question.

There is a new ticket for showing only the tuition types available in the searched for postcode: https://dfedigital.atlassian.net/browse/NTP-451

iterative-it: Nice one. Reed Tutors /tuition-partner/reed-tutors are the best example of a TP that varies price by subject

iterative-it (approved):

PR #72, opened 2022-07-27
NTP-439: Upgrade to GovUk.Frontend.AspNetCore v1.0.0-beta1

iterative-it:
## Context

The service's UI is heavily reliant on the GovUk.Frontend.AspNetCore NuGet package which recently released a [v1.0.0-beta1 version](https://github.com/gunndabad/govuk-frontend-aspnetcore/releases/tag/v1.0.0-beta1). As we want to keep our referenced packages up to date, it is important that we move to this version as soon as possible.

The intention is that there are no visible UI changes as a result of this update.

This change attempts to take advantage of a number of new features in the updated package:

- Pages with validation errors should prepend ["Error: " to the title](https://design-system.service.gov.uk/patterns/validation/). This now happens automatically with the updated library.
- Automatic error summary where supported. It could not be used on the subjects page as it caused a repetition of the error when more than one key stage was selected
- using new govuk-button-link

I also took the opportunity to improve the homepage options cards to get the spacing closer to the design and improve the mobile view.

## Guidance to review

I could not find a better way than the `Index.OnGet` method to resolve the previously invisible validation error on get.

## Link to Jira ticket

[NTP-439](https://dfedigital.atlassian.net/browse/NTP-439)

## Things to check

- [X] Title is in the format "NTP-XXX: Short description" where NTP-XXX is the Jira ticket reference
- [X] Code and tests follow the [coding standards](/docs/coding-standards.md)
- [X] `dotnet format` has been run in the repository root
- [ ] Test coverage of new code is at least 80% - N/A
- [ ] All UI related acceptance criteria specified in the ticket have been added to the relevant [cypress test feature file(s)](/UI/cypress/e2e/) - N/A
- [X] Unless explicitly mentioned in the ticket, all UI work should have been tested as working on desktop and mobile resolutions with JavaScript on and off
- [ ] Database migrations can be applied to the current codebase in main without causing exceptions - N/A
- [ ] Debug logging has been added after all logic decision points - N/A
- [X] All [automated testing](/README.md#testing) including accessibility and security has been run against your code

### DIFF
--- a/UI/Pages/Index.cshtml.cs
+++ b/UI/Pages/Index.cshtml.cs
@@ -24,6 +24,16 @@ public partial class Index : PageModel
 
     public record Command : SearchModel, IRequest<Domain.IResult> { }
 
+    public IActionResult OnGet()
+    {
+        // The model is being validated on get due to [BindProperty(SupportsGet = true)].
+        // This is not appropriate for this page either when first arriving on this page or using a back link.
+        // Therefore clear any model state validation errors
+        ModelState.Clear();
+
+        return Page();
+    }
+
     public async Task<IActionResult> OnPost()
     {
         if (!ModelState.IsValid) return Page();
--- a/UI/Pages/TuitionPartner.cshtml
+++ b/UI/Pages/TuitionPartner.cshtml
@@ -5,9 +5,9 @@
 }
 <a asp-page="@nameof(Index)" class="govuk-link" data-testid="home-link">Home</a><br/>
 
-<govuk-back-link href="/search-results?@Model.AllSearchData.ToQueryString()">
+<a href="/search-results?@Model.AllSearchData.ToQueryString()" class="govuk-back-link">
 	Back to search results
-</govuk-back-link> 
+</a>
 
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
--- a/Application/Application.csproj
+++ b/Application/Application.csproj
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="MediatR.Extensions.Microsoft.DependencyInjection" Version="10.0.1" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="6.0.6" />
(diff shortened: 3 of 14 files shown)


vipinsamreddy (approved):

stephenjmcneill (approved):

UI/Pages/TuitionPartner.cshtml
```diff
@@ -5,9 +5,9 @@
 }
 <a asp-page="@nameof(Index)" class="govuk-link" data-testid="home-link">Home</a><br/>
 
-<govuk-back-link href="/search-results?@Model.AllSearchData.ToQueryString()">
+<a href="/search-results?@Model.AllSearchData.ToQueryString()" class="govuk-back-link">
```
benagain, line 8: I'm curious, what was going wrong when using the tag helper?

UI/Pages/Index.cshtml.cs
```diff
@@ -24,6 +24,16 @@ public Index(ILogger<Index> logger, IMediator mediator)
 
     public record Command : SearchModel, IRequest<Domain.IResult> { }
 
+    public IActionResult OnGet()
```
benagain, line 27: This seems reasonable to me, we want the postcode to be optional for `GET` but mandatory for `POST`.  

I suppose we could do away with the `Data` property and take the postcode as a parameter:

    public void OnGet(string? postcode) {}
    public void OnPost(string postcode) {}

However we would then have to create the `Command` for mediator and it all sounds more complicated that simply suppressing the validation in `OnGet`.

benagain (approved):
I like a bit of dependency upgrade.

UI/Pages/TuitionPartner.cshtml
iterative-it: Frustratingly it was HTML encoding rather than URL encoding the href so & became &amp; which broke the back links again.

I had a brief look at the library's source and couldn't see where that was happening. I'll raise it as a post private beta task to look into it and perhaps raise a PR for that library

PR #157, opened 2022-09-09
NTP-592: Removed pagination code as it is not a requirement

stephenjmcneill:
…fault size is 1.  Main code left intact in case pagination is needed in future.

## Context

Removed pagination code as it is not a requirement

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-592

## Things to check


benagain (commented):
Do you think it would be useful to have an automated test that verifies the total number of results is correct?  Either an e2e test or a unit test.  As the original complaint was that results over 50 were counted incorrectly, a unit test would have to add at least 51 entries.

Tests/SearchForResults.cs
```diff
@@ -308,4 +308,51 @@ private async Task SetupTuitionPartner()
...
+                    new() { TuitionTypeId = (int)TuitionTypes.InSchool, SubjectId = Subjects.Id.KeyStage3Maths, GroupSize = 3, HourlyRate = 56.78m },
+                    new() { TuitionTypeId = (int)TuitionTypes.Online, SubjectId = Subjects.Id.KeyStage3Maths, GroupSize = 3, HourlyRate = 56.78m },
+                }
+                });
+            }
+            await db.SaveChangesAsync();
+        });
+        var result = await Fixture.SendAsync(new SearchResults.Query());
+        result!.Results.Should().NotBeNull();
+        result!.Results!.Count.Should().Be(54);
```
benagain, line 356: `numberOfTuitionPartners` is `55`, I think it would be good if we could change the loop so that it added the number of partners specified, and then we could use `numberOfTuitionPartners` in the assertion.  Otherwise people coming to read this in the future will wonder why the search results returns one less than it looks.

```suggestion
        result!.Results!.Count.Should().Be(numberOfTuitionPartners);
```

UI/cypress/e2e/results.js
```diff
@@ -188,3 +188,17 @@ Then("results are updated after filters are cleared", () => {
...
+Then(
+  "the number of tuition partners displayed matches the displayed count",
+  () => {
+    let countOfElements = 0;
+    cy.get('[data-testid="results-list-item"]').then(($elements) => {
+      countOfElements = $elements.length;
+      cy.get('[data-testid="result-count"]')
+        .invoke("text")
+        .then(parseInt)
+        .should("equal", countOfElements);
```
benagain, line 201: This is a really good way of asserting the count matches the number of results.  Much better than my idea of hard coding the count! 💯

Tests/SearchForResults.cs
stephenjmcneill: Changed

benagain (approved):

PR #166, opened 2022-09-20
NTP-568: add functional tests for cookies code behind

stephenjmcneill:
## Context

Adding unit tests to cookie page

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-568

## Things to check


Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
+﻿using System.Net.Http.Headers;
```
iterative-it, line 1: Don't think this import is used

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
```
iterative-it, line 7: Don't think this import is used

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+using Moq;
```
iterative-it, line 11: Don't think this import is used

Tests/Tests.csproj
```diff
@@ -18,9 +18,11 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="6.7.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
```
iterative-it, line 21: This file can be reverted as this package isn't used

Tests/Tests.csproj
```diff
@@ -18,9 +18,11 @@
...
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="6.0.8" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.1" />
+    <PackageReference Include="Moq" Version="4.18.2" />
```
iterative-it, line 25: This file can be reverted as this package isn't used

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+using Cookies = UI.Pages.Cookies;
+
+
```
iterative-it, line 14: Remove these two extra line breaks

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+    public CookiesTests(SliceFixture fixture) : base(fixture)
+    {
+    }
+
+    [Theory]
+    [InlineData(true, true)]
+    [InlineData(null, true)]
+    [InlineData(true, null)]
+    [InlineData(null, null)]
+    public void Cookies_OnGet_Test(bool? consent, bool? preferencesSet)
```
iterative-it, line 30: It's not clear from the test name or the asserts what this test is checking. It's possible the other tests in this file cover the full logic. If this is testing a specific path, consider renaming it and/or asserting (page) state rather than the result type

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+
+    [Fact]
+    public async void Cookies_For_FindATuitionPartner_Retrieved_OnGet()
+    {
+        var cookiePage = await Fixture.GetPage<Cookies>()
+              .Execute(page =>
+              {
+                  page.PageContext = GetPageContext();
+                  page.Request.Cookies = MockRequestCookieCollection(".FindATuitionPartner", "true");
+                  page.OnGet(true, true);
```
iterative-it, line 66: I think the correct check here is that it uses the cookie value when OnGet(null, <any>) is called. I think you need to check that page.Consent is set to the value from the cookie, not that the cookie is set

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    [Fact]
+    public async void Cookies_For_FindATuitionPartner_Created_OnGet()
+    {
+        var cookiePage = await Fixture.GetPage<Cookies>()
+              .Execute(page =>
+              {
+                  page.PageContext = GetPageContext();
+                  page.OnGet(true, true);
```
iterative-it, line 51: You also need to check that passing consent = false sets the cookie value to false

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+    [Fact]
+    public async void Cookies_Associated_With_FindATuitionPartner_Deleted_When_Set_To_False()
+    {
+        var cookiePage = await Fixture.GetPage<Cookies>()
+              .Execute(page =>
+              {
+                  page.PageContext = GetPageContext();
+                  page.Request.Cookies = MockRequestCookieCollection(".FindATuitionPartner", "true");
+                  page.Request.Cookies.ContainsKey("_ga").Should().BeTrue();
+                  page.Request.Cookies.ContainsKey("_gid").Should().BeTrue();
```
iterative-it, line 120: These lines and some of the lines below are really asserting that your test data is valid. That shouldn't be necessary and adds confusion as to what you're actually testing - the last two lines of this test I think.

Tests/CookiesTests.cs
```diff
@@ -0,0 +1,191 @@
...
+               return Task.FromResult(page);
+           });
+        cookiePage.Should().BeOfType<Cookies>();
+        cookiePage.ModelState.IsValid.Should().BeFalse();
+
+        cookiePage.ModelState.AddModelError("cookies", "You must select an option");
+        var result = cookiePage.OnPost();
+        result.Should().BeOfType<PageResult>();
+    }
+
```
iterative-it, line 170: I don't see direct tests for OnGet preferencesSet or ReturnUrl. There's logic there for when to set it and it should be possible to check the value of ReturnUrl on the page following the call.

iterative-it (changes requested):
There are a lot of changes requested here I know but please don't take that as significant criticism. This is the first major functional test suite you've written in this codebase so it's a good learning experience and worth getting right.

Tests/CookiesTests.cs
stephenjmcneill: Removed

Tests/CookiesTests.cs
stephenjmcneill: Had to use this to mock tempdata setup

Tests/CookiesTests.cs
stephenjmcneill: Removed

Tests/CookiesTests.cs
stephenjmcneill: Removed test individual tests now used

Tests/CookiesTests.cs
stephenjmcneill: The setting of false and true via the consent variable is done other tests

Tests/CookiesTests.cs
stephenjmcneill: Changed as asked

Tests/CookiesTests.cs
stephenjmcneill: Removed

Tests/CookiesTests.cs
stephenjmcneill: Added tests for OnGet preferencesSet or ReturnUrl.

Tests/Tests.csproj
stephenjmcneill: Removed

Tests/Tests.csproj
stephenjmcneill: Being used now

iterative-it (approved):

PR #180, opened 2022-10-11
NTP-621: import gias data during data import process

stephenjmcneill:
## Context

Added code to import GIAS data on schools

Add cs files to import data
Add migration for Entity framework changes to persist import

## Guidance to review
No front end changes can be seen anyone testing will have to debug to prove data has been saved and retrieved.
Information is logged out in console

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-621

## Things to check

- [x] Database migrations can be applied to the current codebase in main without causing exceptions


Application.Tests/Mappings/SchoolDatumTests.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Application.Mapping;
+using Domain.Constants;
+using FluentAssertions;
+using Xunit;
+
+namespace Application.Tests.Mappings;
+public class SchoolDatumTests
+{
+    [Theory]
+    [InlineData(26)]
```
iterative-it, line 10: I think it's worth having these as constants as in making a start on building up the EstablishmentType list so that it's more clear what the test is asserting. This should be used in the `IsValidForService()` method which would allow the comment there to be removed.

Domain.Tests/Validators/SchoolValidatorTests.cs
```diff
@@ -0,0 +1,186 @@
...
+    [InlineData(9998)]
+    public void With_invalid_urn(int id)
+    {
+        var model = new School { Urn = id };
+        var result = _validator.TestValidate(model);
+        result.ShouldHaveValidationErrorFor(x => x.Urn);
+    }
+
+    [Theory]
+    [InlineData(1000)]
```
iterative-it, line 39: I think this test should have failed. Looks like you're accidentally testing LocalAuthorityDistrictId rather than URN

Domain.Tests/Validators/SchoolValidatorTests.cs
```diff
@@ -0,0 +1,186 @@
...
+    [InlineData(null)]
+    public void With_no_address(string address)
+    {
+        var model = new School { Address = address };
+        var result = _validator.TestValidate(model);
+        result.ShouldHaveValidationErrorFor(x => x.Address);
+    }
+
+    [Theory]
+    [InlineData(0)]
```
iterative-it, line 59: Can these be changed to use the enums so it's clear what the integers are? Same for other similar tests below

Domain.Tests/Validators/SchoolValidatorTests.cs
```diff
@@ -0,0 +1,186 @@
...
+    [InlineData(-1)]
+    public void With_invalid_establishmentTypeGroupId(int id)
+    {
+        var model = new School { EstablishmentTypeGroupId = id };
+        var result = _validator.TestValidate(model);
+        result.ShouldHaveValidationErrorFor(x => x.EstablishmentTypeGroupId);
+    }
+
+    [Theory]
+    [InlineData(1)]
```
iterative-it, line 70: You can use something like `Enum.GetValues((typeof(EstablishmentTypeGroup)))` to loop through all the enum values rather than hard coding them like this. That way any new ones will automatically be tested. Same for other tests below

Domain.Tests/Validators/SchoolValidatorTests.cs
```diff
@@ -0,0 +1,186 @@
...
+    public void With_invalid_local_authority_id(int id)
+    {
+        var model = new School { LocalAuthorityId = id };
+        var result = _validator.TestValidate(model);
+        result.ShouldHaveValidationErrorFor(x => x.LocalAuthorityId);
+    }
+
+    [Theory]
+    [InlineData(1)]
+    [InlineData(100)]
```
iterative-it, line 148: The validator itself only checks for greater than zero so I don't think this 100 is needed. It implies there's something significant about 100. Same for lad id below

Domain/Validators/SchoolValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+        RuleFor(m => m.LocalAuthorityDistrictId)
+            .GreaterThan(0)
+            .WithMessage("Enter a valid Local Authority District id");
+
+        RuleFor(m => m.Postcode)
+            .NotEmpty()
+            .WithMessage("Enter a valid Postcode");
+    }
+
+    private bool ValidEstablishmentStatusId(int EstablishmentStatusId)
```
iterative-it, line 48: There's a built in `IsInEnum` validator in fluent validation that I think can be used for these properties once you cast them to the relevant enum https://docs.fluentvalidation.net/en/latest/built-in-validators.html#enum-validator

Infrastructure/EntityConfigurations/EstablishmentStatusConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Infrastructure.EntityConfigurations;
+
+public class EstablishmentStatusConfiguration : IEntityTypeConfiguration<EstablishmentStatus>
+{
+    public void Configure(EntityTypeBuilder<EstablishmentStatus> builder)
+    {
+        builder.HasData(
```
iterative-it, line 12: Missing the somewhat pointless index on the name property

iterative-it (changes requested):
This is a great addition. Having the GIAS dataset available in our service will open the doors to many enhancements and improvements to metrics.

I think all the comments are worth addressing but the only ones that are needed are the `With_valid_urn` test and trying to use constants and enums rather than hard coded integers.

Application.Tests/Mappings/SchoolDatumTests.cs
stephenjmcneill: Changed as requested

Domain.Tests/Validators/SchoolValidatorTests.cs
stephenjmcneill: Cannot get this to work with inbuilt functions due to the to the values not being specified as enums

Domain.Tests/Validators/SchoolValidatorTests.cs
stephenjmcneill: Removed

Domain/Validators/SchoolValidator.cs
stephenjmcneill: Cannot get this to work with inbuilt functions due to the to the values not being specified as enums

Infrastructure/EntityConfigurations/EstablishmentStatusConfiguration.cs
stephenjmcneill: Added

iterative-it (approved):

PR #187, opened 2022-11-09
NTP-659: pricing info not displaying for t ps when do you offer tuition to all locations nationwide response is empty

kieranohanlon:
## Context

When a search is done the postcode (and LADs) are stored in temp data/session and then used to filter the details that are shown on the TP details page.  The same page is also called from the full list of TPs, which should show nationwide data.  The issue is that it's still applying the previously searched postcode/LADs that are in temp data.  

The suggestion is to switch this to using the query string that is similar to the rest of the website.

To reproduce the issue do the following…

Do a search, with the following details

image

If you then search for “Action Tutoring” on this page it will not be found, since it does not supply tutoring for this postcode.

Return to the full list of tutors and do the following

image

Selecting the “Action Tutoring” TP shows the following, with the type of tuition and price details missing (since it applies the previously set postcode filter to the data, but shouldn’t)

image

The data should be displayed as follows…

image

## Guidance to review

The end-to-end tests use live data to verify the tuition types are shown / not shown when appropriate. This is the easiest way to write these tests as we currently have no way to manipulate the back-end data from cypress.

## Link to Jira ticket

[](https://dfedigital.atlassian.net/browse/NTP-659)

## Things to check

### DIFF
--- a/Application/INtpDbContext.cs
+++ b/Application/INtpDbContext.cs
@@ -6,6 +6,7 @@ namespace Application;
 public interface INtpDbContext
 {
     DbSet<LocalAuthorityDistrict> LocalAuthorityDistricts { get; set; }
+    DbSet<LocalAuthorityDistrictCoverage> LocalAuthorityDistrictCoverage { get; set; }
     DbSet<Region> Regions { get; set; }
     DbSet<Subject> Subjects { get; set; }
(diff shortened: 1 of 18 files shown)


Kolapo-Abolarinwa (approved):

ashward (approved):
lgtm

PR #196, opened 2022-11-24
Ntp 640 comparing and contrasting t ps desktop

kieranohanlon:
## Things to check


Domain/Validators/TuitionPartnersFilterRequestValidator.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Domain.Validators;
+
+public abstract class TuitionPartnersFilterRequestValidator<T>
+{
+    protected TuitionPartnersFilterRequestValidator()
```
ashward, line 5: Do we need this empty constructor?

Infrastructure/MetricLogging/LoggingTuitionPartnerService.cs
```diff
@@ -0,0 +1,75 @@
...
+        _logger.LogInformation("Filtering tuition partners");
+
+        var stopwatch = new Stopwatch();
+        stopwatch.Start();
+        var result = await _inner.GetTuitionPartnersFilteredAsync(filter, cancellationToken);
+        stopwatch.Stop();
+
+        var resultCount = result?.Length;
+
+        var logLevel = (resultCount == 0 && string.IsNullOrWhiteSpace(filter.Name)) ? LogLevel.Warning : LogLevel.Information;
```
ashward, line 29: What's the reason that we might log this case as a warning rather than as info?

UI/Pages/Shortlist.cshtml.cs
```diff
@@ -0,0 +1,172 @@
...
+            };
+
+            static FluentValidationResult UnknownError() =>
+                new(new[] { new ValidationFailure("", "An unknown problem occurred") });
+        }
+
+        private int[] GetShortlist()
+        {
+            //TODO - integrate with shortlist selection
+            var tuitionPartnersIds = new int[] { 124, 125, 126, 127, 129, 131 };
```
ashward, line 104: Using the row ids will mean that shortlists will become invalid after a data import (as the ids all change). I wonder is we should use the seoUrl for the time being as that's the key we use in most other places?

kieranohanlon: Quite possibly, sounds like a good shout - I have just mocked this up for now, but we can switch it to seoUrl when Kola passes me the info over via cookie.

Infrastructure/MetricLogging/LoggingTuitionPartnerService.cs
kieranohanlon: This is lifted from the existing "SearchTuitionPartnerHandlerLoggingBehaviour.cs" - I guess at the moment there should be no filters possible (other than Name) that would return 0 results.  So if they do return 0 it's probably an issue, something has gone wrong.  But if we add any new filters in the future and it results in 0 results on a regular basis then I think we'd want it to switch to info.

Domain/Validators/TuitionPartnersFilterRequestValidator.cs
kieranohanlon: I suspect not - I think I just followed "TuitionPartnerSearchRequestValidator.cs", but think that the constructor can be removed from both.  I'll update that.

UI/Pages/TuitionPartner.cshtml.cs
```diff
@@ -181,27 +167,102 @@ public QueryHandler(ILocationFilterService locationService, INtpDbContext db, IL
...
+            {
+                return tpError.Cast<TuitionPartnersResult>();
+            }
+
+            var result = new TuitionPartnersResult(tpResult.Data, location.LocalAuthority);
+
+            return Result.Success(result);
+        }
+
+        private async Task<IResult<LocationFilterParameters>> GetSearchLocation(Query request, CancellationToken cancellationToken)
```
ashward, line 204: This logic is repeated in a few places. Would be good to refactor into a single place.

kieranohanlon: Yep, I have that on my todo list - possibly move it in to a base handler.  There are a few other bits that I want to refactor - can go through them with you and Kola.  But wanted to get this in to 720 so we start on the merge ASAP.

Domain/Validators/TuitionPartnersFilterRequestValidator.cs
kieranohanlon: Removed these empty constructors.

UI/Pages/Shortlist.cshtml.cs
kieranohanlon: Updated to use SeoURL.

ashward (approved):
lgtm

Kolapo-Abolarinwa (approved):

PR #200, opened 2022-12-01
NTP-647: Add comparing and contrasting Tuition Partners functionality on search result page

Kolapo-Abolarinwa:
## Context

User research has confirmed that users want to compare and contrast a shorter list of TPs in order to help make a more informed decision on which tuition partner can meet their service needs. As such this ticket will add some functionality to the search results page to allow the users to compare and contrast a set number of tuition partners on one page.

- Add the ability to select the different Tuition Partners you would like to compare.
- Have some visual representation that you have selected Tuition Partners.
- Be able to navigate to another page that compares your selected Tuition Partners.
- The above functionality should be performable with or without JavaScript enabled in the browser.

## Guidance to review
on landing on the Search Result Page:
- A 'View my shortlisted tuition partners' should be present and functional 
- A badge showing selected Tuition Partners to compare should be present and functional
- If JavaScript is enabled:
  - 'Update my shortlist' buttons should not be visible.
  - Checkboxes to allow the selection of Tuition Partners should be present and functional:
    + in the checked state it should update the badge correctly
    + in an unchecked state it should update the badge correctly
- If JavaScript is not enabled:
   + Checkboxes to allow the selection of Tuition Partners should be present and functional:
    + two 'Update my shortlist' buttons should be present, one at the top and the other at the bottom of the page.

## Link to Jira ticket

https://dfedigital.atlassian.net/jira/software/projects/NTP/boards/135?selectedIssue=NTP-647

## Things to check

- [ ] All UI-related acceptance criteria specified in the ticket have been added to the relevant [cypress test feature file(s)](/UI/cypress/e2e/)
- [ ] **PR deployment has been signed off by the wider team**


Application/Application.csproj
```diff
@@ -16,11 +16,20 @@
...
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="6.0.8" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\Domain\Domain.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Http.Abstractions">
+      <HintPath>..\..\..\..\..\..\usr\local\share\dotnet\x64\shared\Microsoft.AspNetCore.App\6.0.4\Microsoft.AspNetCore.Http.Abstractions.dll</HintPath>
```
ashward, line 30: What's this path to? This won't be the same for all devs.

Application/StorageManager/Interfaces/IStorageManager.cs
```diff
@@ -0,0 +1,52 @@
+using Domain.Search.ShortlistTuitionPartners;
+
+namespace Application.StorageManager.Interfaces;
+
+public interface IStorageManager
```
ashward, line 5: `IStorageManager` is a pretty generic name. Maybe this should be `ITuitionPartnerShortlistStorage` or something?

Kolapo-Abolarinwa: Yes, you are right. Kieran and I had talked about renaming it. I will do it soonest👍

Application/Application.csproj
Kolapo-Abolarinwa: Thanks, Ash, I noticed it too. I am digging into it.

Application/StorageManager/Implementations/CookieTuitionStorageManager.cs
```diff
@@ -0,0 +1,181 @@
...
+        _httpContextAccessor.HttpContext?.Request.Cookies
+            .Where(c => c.Key.Contains(CookieKeyFormat));
+
+    private TuitionPartner? GetTuitionPartnerFromCookie(string key)
+    {
+        var cookieKey = ConstructCookieKey(key);
+        if (!IsTuitionPartnerPresent(cookieKey)) return null;
+
+        var cookie = _httpContextAccessor.HttpContext?.Request.Cookies[$"{cookieKey}"];
+        var tuitionPartnerDetails = cookie?.Split("&");
```
ashward, line 166: Seems strange that we're having to handle url form encoded values ourselves. Could you use something like the `CookieHeaderValue` class, as per this https://learn.microsoft.com/en-us/aspnet/web-api/overview/advanced/http-cookies#structured-cookie-data ?

Kolapo-Abolarinwa: I definitely agree with the 'CookieHeaderValue' implementation. After looking into it again in more detail. I currently can't find a way to make it work in a class project. When using it in a web API/MVC/Razor pages project, the CookieHeaderValue implementation seems happy. But in a class project, [even after using different NuGet packages] it doesn't seem to work.

Domain/Search/ShortlistTuitionPartners/TuitionPartner.cs
```diff
@@ -0,0 +1,3 @@
+namespace Domain.Search.ShortlistTuitionPartners;
+
+public record TuitionPartner(string SeoUrl, string? LocalAuthority);
```
ashward, line 3: I wonder if this should be something like `ShortlistedTuitionPartner` or something like that, so that it doesn't get confused with an actual `TuitionPartner` record?

Domain/Search/TuitionPartnerSearchResult.cs
```diff
@@ -3,7 +3,7 @@
 public class TuitionPartnerSearchResult
 {
     public int Id { get; set; }
-    public string SeoUrl { get; set; } = null!;
+    public string? SeoUrl { get; set; } = null!;
```
ashward, line 6: Would we have a case where the `SeoUrl` is null?

UI/src/sass/search-result-page-shortlist.scss
```diff
@@ -0,0 +1,25 @@
...
+  color: govuk-colour("black");
+  padding: 0.5rem 0.5rem 0.5rem 0.5rem;
+}
+.my-shortlisted-partners-badge--three-digits-text {
+  background-color: govuk-colour("mid-grey");
+  color: govuk-colour("black");
+  padding: 0.7rem 0.4rem 0.4rem 0.4rem;
+}
+//Overriding the default value
+.govuk-checkboxes--small .govuk-checkboxes__label {
```
ashward, line 23: This CSS selector should be more tightly scoped to just the shortlist checkbox, otherwise it might affect other uses of small checkboxes across the site.

UI/Pages/SearchResults.cshtml.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+    if (badgeElement)
+      badgeElement.textContent = result.totalShortlistedTuitionPartners;
+  } else {
+    checkBox.checked = elseCheckboxState;
+  }
+};
+const onChecked = async (checkbox) => {
+  return await fetch(
+    getAddShortlistTuitionPartnerFetchUrl(
```
ashward, line 35: This should really be a POST rather than a GET

UI/Pages/SearchResults.cshtml.js
```diff
@@ -0,0 +1,82 @@
...
+  )
+    .then((response) => getFetchResponseData(response))
+    .catch((error) => {
+      //What do we want to do with the error.
+      throw Error(error);
+    });
+};
+
+const onUnChecked = async (checkbox) => {
+  return await fetch(getRemoveShortlistTuitionPartnerFetchUrl(checkbox.id))
```
ashward, line 48: This should really be a POST rather than a GET

UI/Pages/SearchResults.cshtml
```diff
@@ -130,45 +158,77 @@
...
+                                        @item.Description
+                                    </govuk-summary-list-row-value>
+                                </govuk-summary-list-row>
+                                <govuk-summary-list-row class="govuk-body-s">
+                                    @{
+                                        var shortlist = Model.SelectableTuitionPartners
+                                            .First(tp => tp.SeoUrl == item.SeoUrl);
+                                    }
+                                    <govuk-summary-list-row-key>
+                                        Add to my shortlist
```
ashward, line 204: This should be the `<label>` for the checkbox

UI/Pages/SearchResults.cshtml
```diff
@@ -130,45 +158,77 @@
...
+                                    </govuk-summary-list-row-key>
+                                    <govuk-summary-list-row-value data-testid="results-description">
+                                        <div class="govuk-checkboxes govuk-checkboxes--small " data-module="govuk-checkboxes">
+                                            <div class="govuk-checkboxes__item shortlist-tuition-partner-checkbox">
+                                                <input class="govuk-checkboxes__input" id="@shortlist.SeoUrl" name="ShortlistedTuitionPartners"
```
ashward, line 209: Could we prefix the `id` attribute here with `shortlist-cb` or something like that?
```suggestion
                                                <input class="govuk-checkboxes__input" id="shortlist-cb-@shortlist.SeoUrl" name="ShortlistedTuitionPartners"
```

UI/Pages/SearchResults.cshtml
Kolapo-Abolarinwa: Yes, it should be. The challenge is, based on the 'dgs' system checkboxes, it expects a label in the same div as the checkbox. And when using that with the `govuk-summary-list-row` element it gets even more interesting.

UI/Pages/SearchResults.cshtml.js
Kolapo-Abolarinwa: Yes, I agree it should be 'POST'. This is what I was looking/saying about the Page, it does everything via 'GET' even things I think should maybe POST. I thought about refining it but the implications may be large so, I decided to stick with the current trend of using GET.

UI/Pages/SearchResults.cshtml.js
Kolapo-Abolarinwa: Yes, I agree it should be 'POST'. This is what I was looking/saying about the Page, it does everything via 'GET' even things I think should maybe POST. I thought about refining it but the implications may be large so, I decided to stick with the current trend of using GET.

Infrastructure/Migrations/20220705165059_CoverageIndexes.Designer.cs
```diff
@@ -4181,7 +4181,7 @@ protected override void BuildTargetModel(ModelBuilder modelBuilder)
 
                     b.Navigation("LocalAuthorityDistrict");
 
-                    b.Navigation("TuitionPartner");
+                    b.Navigation("ShortlistedTuitionPartner");
```
kieranohanlon, line 4184: Still quite a few "ShortlistedTuitionPartner" changes to various Migration Designer files that need reverting

Infrastructure/Migrations/NtpDbContextModelSnapshot.cs
```diff
@@ -4432,7 +4432,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Navigation("LocalAuthorityDistrict");
 
-                    b.Navigation("TuitionPartner");
+                    b.Navigation("ShortlistedTuitionPartner");
```
kieranohanlon, line 4435: Needs reverting?

Domain/Search/TuitionPartnerResult.cs
```diff
@@ -3,7 +3,7 @@
 public class TuitionPartnerResult
 {
     public int Id { get; set; }
-    public string SeoUrl { get; set; } = null!;
+    public string? SeoUrl { get; set; } = null!;
```
kieranohanlon, line 6: Does this need to be nullable?  It's not NULL in the database.

Application/TuitionPartnerShortlistStorage/Implementations/CookieBasedTuitionPartnerShortlistStorage.cs
```diff
@@ -0,0 +1,140 @@
+namespace Application.TuitionPartnerShortlistStorage.Implementations;
```
ashward, line 1: I think we should move this into the UI project so that we can use the `CookieHeaderValue` stuff.

PR #201, opened 2022-12-01
NTP-717: Ensure checkboxes can be selected on filters with no js

kieranohanlon:
## Context

To reproduce:

Disable JS

Carry out a search and get to the results page

Attempt to change the subject checkboxes on the left side

Expected result:

At 3. The checkboxes can be changed and ‘Apply filters’ will update the results

Actual result:

At 3. The checkboxes cannot be clicked with the mouse (although they can be selected with the keyboard).

The issue is that the "Key Stage" filter headings (that collapse and expand the sections when js is enabled) is set as a an HTML button, which is correct when js is enabled, but by design it should be a span when js is disabled.   This then allows the checkboxes to be selected.  

The following component is used and details about this are on line 131 of the js file:
https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/components/option-select.js 

This work was originally done as part of 
https://dfedigital.atlassian.net/browse/NTP-355

## Guidance to review

Follow the steps above and test without js

## Link to Jira ticket

[](https://dfedigital.atlassian.net/browse/NTP-717)

## Things to check

### DIFF
--- a/UI/Pages/Shared/_OptionsSelect.cshtml
+++ b/UI/Pages/Shared/_OptionsSelect.cshtml
@@ -12,7 +12,7 @@
         <div class="app-c-option-select__container-inner">
             <div class="gem-c-checkboxes govuk-form-group govuk-checkboxes--small" data-module="gem-checkboxes">
                 <fieldset class="govuk-fieldset govuk-checkboxes--small govuk-!-margin-left-1">
-                    <legend class="govuk-fieldset__legend govuk-fieldset__legend--m govuk-visually-hidden">@Model.DisplayName</legend>
+                    <legend class="govuk-fieldset__legend govuk-fieldset__legend--s app-c-option-select__legend">@Model.DisplayName</legend>
 
                     <ul class="govuk-checkboxes gem-c-checkboxes__list" data-module="govuk-checkboxes">
                         @foreach (var subject in Model.Items)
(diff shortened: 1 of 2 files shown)


UI/Pages/Shared/_OptionsSelect.cshtml
```diff
@@ -3,10 +3,7 @@
 <div class="app-c-option-select js-collapsible" data-testid="@Model.Name" data-module="option-select" @Model.ClosedData>
 
     <div class="app-c-option-select__heading js-container-heading">
-        <button class="js-container-button app-c-option-select__title app-c-option-select__button" type="button"
-                id="option-select-title-@Model.Name" aria-controls="list-of-@Model.Name">
-            @Model.DisplayName
-        </button>
+        <span class="app-c-option-select__title js-container-button" id="option-select-title-@Model.Name">@Model.DisplayName</span>
```
ashward, line 6: Could we use the existing `<legend>` from the `<fieldset>` for no-JS so that it looks the same as the 'Type of tuition'? Or move the existing legend to here, and set it as visually hidden is JS is enabled maybe?

kieranohanlon: Made it so that it shows the button if js is enabled, but shows legend (updated to "govuk-fieldset__legend--s") if not enabled.

UI/src/sass/option-select.scss
```diff
@@ -116,10 +116,19 @@ ul.gem-c-checkboxes__list {
...
+
 .js-enabled {
   .app-c-option-select__heading {
     position: relative;
     padding: 10px 8px 0px 35px;
+    display: block;
+  }
+
+  .app-c-option-select__legend {
+    display: none;
```
ashward, line 131: Could we use the `govuk-visually-hidden` mixin here so that it's still available to screen readers?

kieranohanlon: Good shout - I have updated this to extend the .govuk-visually-hidden class

ashward (approved):
lgtm

PR #214, opened 2022-12-16
NTP-756: Add/Remove Tuition Partner to/from shortlist

Kolapo-Abolarinwa:
## Context
From the Tuition Partner [TP] information page, we need a way to add and remove a TP to/from the shortlist, 
when the user accesses the TP information page either from the TP link on the shortlist TP page 
or from the search results page. 
The  Search Result page should also be updated as required with a TP is added to or removed from 
the shortlist via the TP information page. 

Implement adding/removing a TP via the TP's information page [using checkboxes].
The information displayed on the page should reflect if the TP has been added or removed from the shortlist.
NB: The solution should work with Javascript [JS] enabled or disabled

## Guidance to review

One should be able to add or remove a TP from the shortlist via the TP's information page regardless of 
if JS is enabled or disabled.
The information displayed on the TP's information page should reflect if the TP has been added to the shortlist.
The Search Result page should reflect the addition or removal of a TP to the shortlist via the TP's information page.

## Link to Jira ticket

[NTP-756](https://dfedigital.atlassian.net/jira/software/projects/NTP/boards/135?selectedIssue=NTP-756)

## Things to check

- [X] Title is in the format "NTP-XXX: Short description" where NTP-XXX is the Jira ticket reference
- [X] `dotnet format` has been run in the repository root
- [X] Test coverage of new code is at least 80%

[NTP-756]: https://dfedigital.atlassian.net/browse/NTP-756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


UI/Pages/TuitionPartner.cshtml
```diff
@@ -4,260 +4,289 @@
...
+
+    <p class="govuk-body">
+        Each pupil premium eligible student will receive £162. This funding cost has been calculated so you will have 60% of the cost covered for up to £18 per pupil per hour. For help understanding your funding, you can see our information on <a href="/funding-and-reporting" class="govuk-link" data-testid="funding-guidance-1">funding and reporting</a>.
+    </p>
+
+    <div>
+        <govuk-details data-testid="payment-details">
             <govuk-details-summary>
-                What is a quality assured tuition partner?
+                <govuk-details-text>
```
kieranohanlon, line 124: I don't this is needed?  The existing markup already appears to render as https://design-system.service.gov.uk/components/details/

UI/MediatR/Handlers/GetTuitionPartnerQueryHandler.cs
```diff
@@ -0,0 +1,273 @@
+using Domain;
```
kieranohanlon, line 1: As we talked over before - think this is a good idea, but think we need to make sure we do the rest of the site in a similar way.  Worth adding a new tech debt ticket to the board for this?

UI/Pages/TuitionPartner.cshtml.js
```diff
@@ -0,0 +1,73 @@
...
+  return await fetch(
+    getAddToShortlistFetchUrl(),
+    getAdditionalFetchParameter(checkboxId)
+  )
+    .then((response) => getFetchResponseData(response))
+    .catch((error) => {
+      throw Error(error);
+    });
+};
+const onUnChecked = async (checkboxId) => {
```
kieranohanlon, line 37: As mentioned to you directly... should the "govuk-caption-l" be updated on check/uncheck to show the right caption like when the page loads?  TBC.

UI/Services/TuitionPartnerShortlistStorage/CookieBasedTuitionPartnerShortlistStorage.cs
```diff
@@ -84,6 +84,16 @@ public void RemoveTuitionPartner(string shortlistedTuitionPartnerSeoUrl)
     public void RemoveAllTuitionPartners() =>
         _httpContextAccessor.HttpContext?.Response.Cookies.Delete(CookieName);
 
+    /// <inheritdoc/>
+    public bool IsTuitionPartnerShortlisted(string tuitionPartnerSeoUrl)
```
kieranohanlon, line 88: Issue when testing the js disabled version...
If you go in to a TP detail, then check the checkbox (if it's not already checked) and keep pressing the "update shortlist" button over and over then it adds multiple versions of the TP to the cookie.  Meaning that the number of shortlisted TPs goes up when it shouldn't.
I think that the CookieBasedTuitionPartnerShortlistStorage.AddTuitionPartnerToCookie (and maybe AddTuitionPartnersToCookie) should only add the TP if it doesn't already exist?

UI/Pages/TuitionPartner.cshtml.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+};
+const onCheckboxClick = async (event) => {
+  const checkbox = event.target;
+  if (checkbox.checked) {
+    try {
+      const result = await onChecked(checkbox.id);
+      if (!isResultValid(result)) throw getError();
+    } catch (error) {
+      //What do we want to do with error?
```
kieranohanlon, line 54: Maybe we should talk about this as a dev team?  Possibly go to error page?  What happens for other failed js get/post calls?  I think we should ensure all do the same.

UI/cypress/e2e/shortlist.feature
```diff
@@ -80,7 +80,7 @@ Feature: Tuition Partner shortlist
         And they choose to view their shortlist from the results page
         When they choose to view the 'Action Tutoring' details from the shortlist
         Then the back link's text is 'Back to shortlist'
-        And the heading caption is 'Tuition partner for Stockport'
+        And the heading caption is 'Shortlisted tuition partner for Stockport'
```
kieranohanlon, line 83: Do we not need some new cypress tests for this PR?

UI/Pages/TuitionPartner.cshtml
Kolapo-Abolarinwa: Okay, on Rider I believe it flags a warning about it which was why I thought to change it.

UI/Pages/TuitionPartner.cshtml
```diff
@@ -4,260 +4,289 @@
...
+    }
 </div>
+</div>
+<div show-if="@Model.Data.HasLogo" class="govuk-grid-column-one-third tuition-partner-logo">
+    <img src="/tuition-partner-logo/@Model.Data.Id" alt="The company logo for @Model.Data.Name" data-testid="tuition-partner-website-logo"/>
+</div>
+</div>
+</form>
+
+@section Scripts{<script src="/Pages/TuitionPartner.cshtml.js"></script>}
```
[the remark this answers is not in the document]
Kolapo-Abolarinwa: I could change it. However, it makes it a lot more complicated to write targeted Javascript. I guess I wonder if the wins match the effort.

UI/MediatR/Handlers/GetTuitionPartnerQueryHandler.cs
Kolapo-Abolarinwa: Yes, I do agree. My thinking was given this is possibly one of the simpler pages why not start from this one?

UI/cypress/e2e/shortlist.feature
Kolapo-Abolarinwa: I will look into this👍

UI/Pages/TuitionPartner.cshtml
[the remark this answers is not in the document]
kieranohanlon: Think it's worth a dev team chat on this one next week?

UI/MediatR/Handlers/GetTuitionPartnerQueryHandler.cs
kieranohanlon: One to cover in the dev team chat and make sure it's as a tech debt ticket on the board then?

UI/Pages/TuitionPartner.cshtml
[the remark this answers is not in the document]
kieranohanlon: Added ne Jira ticket 838 for this

UI/MediatR/Handlers/GetTuitionPartnerQueryHandler.cs
kieranohanlon: Added new ticket in Jira - 839

UI/Pages/TuitionPartner.cshtml.js
kieranohanlon: Added a new Jira ticket to deal with the way we handle js errors - #841

Kolapo-Abolarinwa:
I think you are right. I was working under the impression no
duplicates should ever need to be passed in.
I would change it to pick just distinct entries in '
AddTuitionPartnersToCookie'👍

UI/Services/TuitionPartnerShortlistStorage/CookieBasedTuitionPartnerShortlistStorage.cs
Kolapo-Abolarinwa: I think you are right. I was working under the impression no duplicates should ever need to be passed in.
I would change it to pick just distinct entries in `AddTuitionPartnersToCookie` 👍

kieranohanlon (approved):

kieranohanlon (approved):

UI/cypress/e2e/shortlist.js
```diff
@@ -144,3 +143,39 @@ When(
     cy.wait(500);
   }
 );
+
+Then("total amount of shortlisted TPs is {int}", (expectedTotal) => {
```
ashward, line 147: Isn't this the same as `"the shortlist shows as having {int} entries on the results page"` above?

UI/cypress/e2e/shortlist.js
```diff
@@ -144,3 +143,39 @@ When(
...
+  cy.checkTotalTps(expectedTotal);
```
ashward, line 148: This is quite a specific thing, but with a generic name, on the global object. Not sure we need any of these on the global object any more do we?

PR #216, opened 2023-01-04
NTP 774: Update spreadsheet importer to be able to import the Tribal report format

kieranohanlon:
## Context

Update spreadsheet importer to be able to import the Tribal report format

## Changes proposed in this pull request

Allow the Tribal spreadsheet format to be used alongside the existing spreadsheet format when importing data from the Google drive.

## Guidance to review

Review the code and run the import to ensure it works as expected.

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-774

## Things to check

- [x] Title is in the format "NTP-XXX: Short description" where NTP-XXX is the Jira ticket reference
- [x] Code and tests follow the [coding standards](/docs/coding-standards.md)
- [x] `dotnet format` has been run in the repository root
- [x] Test coverage of new code is at least 80%
- [ ] All UI related acceptance criteria specified in the ticket have been added to the relevant [cypress test feature file(s)](/UI/cypress/e2e/)
- [ ] Unless explicitly mentioned in the ticket, all UI work should have been tested as working on desktop and mobile resolutions with JavaScript on and off
- [x] Database migrations can be applied to the current codebase in main without causing exceptions
- [x] Debug logging has been added after all logic decision points
- [x] All [automated testing](/README.md#testing) including accessibility and security has been run against your code
- [ ] **PR deployment has been signed off by wider team**

PR #220, opened 2023-01-12
NTP-810: Filter by organisation type

kieranohanlon:
## Context

Filter by organisation type

Ensure that the organisation type is collected in a more structured way (stored as a list of values in a new table with a FK on the main TP table), rather than as a string.  This will allow us to then build a radio button filter that will appear on the search results below the existing Tuition Type filter.

## Guidance to review

Ensure that the TPs match the list of agreed organisation types and then ensure that the functionality works.  Test with and without js enabled, on mobile and check accessibility

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-810

## Things to check


Infrastructure/Factories/QualityAssuredSpreadsheetTuitionPartnerFactory.cs
```diff
@@ -76,12 +79,19 @@ public async Task<TuitionPartner> GetTuitionPartner(Stream stream, CancellationT
...
             Experience = _spreadsheetExtractor.GetCellValue(GeneralInformationSheetName, "C", 13),
-            LegalStatus = _spreadsheetExtractor.GetCellValue(GeneralInformationSheetName, "D", 13),
             HasSenProvision = _spreadsheetExtractor.GetCellValue(GeneralInformationSheetName, "F", 13).ParseBoolean(),
             AdditionalServiceOfferings = _spreadsheetExtractor.GetCellValue(GeneralInformationSheetName, "G", 13),
             IsVatCharged = _spreadsheetExtractor.GetCellValue(PricingSheetName, "K", 2).ParseBoolean(),
         };
 
+        //Get Organisation Type Id
+        var organisationTypeName = _spreadsheetExtractor.GetCellValue(GeneralInformationSheetName, "D", 13);
+        var organisationType = _organisationTypes!.Where(x => string.Equals(x.Name, organisationTypeName, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
```
hasan3d, line 89: It can be simplified and replaced with a single call to FirstOrDefault.
`var organisationType = _organisationTypes!.FirstOrDefault(x => string.Equals(x.Name, organisationTypeName, StringComparison.InvariantCultureIgnoreCase));`

If we expect a unique value for the organisation type name, we can use `SingleOrDefault` instead of `FirstOrDefault` to enforce that.

UI/Enums/OrganisationTypeGrouping.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.ComponentModel;
+
+namespace UI.Enums
```
hasan3d, line 3: The enum file should be in the `Domain/Enums` folder instead of here.

hasan3d (commented):
I have now concluded my review. The code looks good overall. So, well done! However, I have left some comments on the PR, but nothing major if you can address those. Thanks!

Domain/Validators/TuitionPartnerValidator.cs
```diff
@@ -61,4 +66,9 @@ private static bool CheckRegex(string property, string regex)
                          regex,
                          RegexOptions.IgnoreCase);
     }
+
+    private bool ValidOrganisationTypeId(int OrganisationTypeId)
```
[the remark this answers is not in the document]
kieranohanlon: Good spot

UI/Enums/OrganisationTypeGrouping.cs
kieranohanlon: Agreed, think I missed this because I just copied the existing TuitionType filter which has it there.  But you'll notice in https://github.com/DFE-Digital/find-a-tuition-partner/pull/216 that I have already moved TuitionType to Domain/Enums.  So I'll move OrganisationTypeGrouping and TuitionType and others will be updated as part of PR 216

hasan3d: Okay, that's fine.

hasan3d (approved):
It looks good now.

kieranohanlon:
There was a decision not to push this live.

PR #224, opened 2023-01-19
NTP-798: Fix a bug with the Invalid or Blank PostCode search on the results page

hasan3d:
## Context

Fix a bug with the Invalid or Blank PostCode search on the results page.

This bug fix also covers the NTP-797: Shortlist page logs an error if the postcode is missing.

Ensure that when a search is carried out with a blank or invalid postcode on the 'search results page, the postcode validator validates the postcode and displays the appropriate validation error message.

Ensure on the 'shortlist page' If the postcode is blank or invalid, and then we send the user straight back to the 'search results page; the postcode validator validates the postcode and displays the appropriate validation error message.

## Guidance to review

Ensure that the postcode validation works on the 'search results' and 'shortlist' pages, and with a blank or invalid postcode, ensure the functionality works. Test with and without js enabled on mobile and check accessibility.

## Link to Jira ticket

[https://dfedigital.atlassian.net/browse/NTP-798](https://dfedigital.atlassian.net/browse/NTP-798)

## Things to check

### DIFF
--- a/Tests/SearchForResults.cs
+++ b/Tests/SearchForResults.cs
@@ -177,7 +177,7 @@ public class SearchForResults : CleanSliceFixture
     [Fact]
     public async Task Displays_all_tutor_types_in_database()
     {
-        // Given
+        // Arrange
         await Fixture.AddTuitionPartner(A.TuitionPartner
             .WithName("a", "Alpha")
             .TaughtIn(District.Dacorum, TuitionTypes.InSchool)
@@ -185,11 +185,11 @@ public class SearchForResults : CleanSliceFixture
                 .Subject(Subjects.Id.KeyStage1English, l => l
                     .InSchool().Costing(12m).ForGroupSizes(2))));
 
-        // When
+        // Act
         var result = await Fixture.SendAsync(
             Basic.SearchResultsQuery with { Postcode = District.Dacorum.SamplePostcode });
 
-        // Then
+        // Assert
         result.Results.Should().NotBeNull();
         result.Results!.Results.Should().BeEquivalentTo(new[]
         {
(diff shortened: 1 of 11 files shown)


Tests/SearchForResults.cs
```diff
@@ -223,43 +229,82 @@ public async Task Preserves_selected_from_querystring()
...
             });
+
+        result.Postcode.Should().Be(District.Dacorum.SamplePostcode);
     }
 
     [Fact]
-    public async Task Verify_Search_Results_Count()
+    public async Task Verify_Search_Results_Count_With_Valid_PostCode()
     {
+        // Arrange
```
kieranohanlon, line 239: Minor thing - don't mind moving these to be named as AAA, rather than GWT.  But have all the unit tests been renamed so it's consistent across the solution?

UI/Pages/Shortlist.cshtml.cs
```diff
@@ -158,7 +163,7 @@ private string[] GetShortlistSeoUrls()
             if (location is IErrorResult error)
             {
                 //Shouldn't be invalid, unless query string edited - since postcode on this page comes from previous page with validation
-                _logger.LogWarning("Invalid postcode '{Postcode}' provided on shortlist page", request.Postcode);
+                _logger.LogInformation("Invalid postcode '{Postcode}' provided on shortlist page", request.Postcode);
```
kieranohanlon, line 166: Guess that the validation kicks in before this?  So should never get here now anyway?  If that's the case, maybe leave it as a warning, so we see the log if it does somehow get to this point?

hasan3d: Yeah, good point. With the new validation logic in place, we validate input before calling the `Data = await _mediator.Send(data);` therefore, we can safely revert that logic to its previous log level, which is a warning.

Tests/SearchForResults.cs
hasan3d: Agree; I will make changes to the rest of the files where we have `GWT (Given, When, Then)` naming in the unit test instead of the `AAA (Arrange, Act, Assert)` to make it consistent across the solution.

kieranohanlon (approved):
Looks good to me

PR #228, opened 2023-01-25
NTP-863: Change log level from warning to info when no Tuition Partner is found

hasan3d:
## Context

Currently, when someone passes an invalid tuition partner id not found on our system, we log a warning that creates an alert for us to look into that. As the tuition partner id is not found, an error classed as 404 error, we will log that as info instead of a warning that will not trigger an alert.

Change log level from warning to information when there is no TP found.

## Guidance to review

Ensure that when no Tuition Partner is found for a given id on the`TuitionPartner.cshtml.cs`, the error is logged as information, and we do not get an alert for that error.

## Link to Jira ticket

[https://dfedigital.atlassian.net/browse/NTP-863](https://dfedigital.atlassian.net/browse/NTP-863)

## Things to check

### DIFF
--- a/UI/Pages/TuitionPartner.cshtml.cs
+++ b/UI/Pages/TuitionPartner.cshtml.cs
@@ -21,7 +21,7 @@ public class TuitionPartner : PageModel
         SearchModel = new SearchModel(query);
 
         if (string.IsNullOrWhiteSpace(query.Id))
-            return ReturnNotFound($"Null or whitespace id '{query.Id}' provided");
+            return ReturnNotFoundWithLogging($"Null or whitespace id '{query.Id}' provided", LogLevel.Information);
 
         Data = await _mediator.Send(query);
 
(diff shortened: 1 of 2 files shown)


UI/Pages/TuitionPartner.cshtml.cs
```diff
@@ -79,9 +79,9 @@ public async Task<IActionResult> OnPostRemoveFromShortlist([FromBody] string seo
     private ArgumentException GetArgumentException(string name) => new($"{name} is null or whitespace");
     private JsonResult GetShortlistJsonResult(bool status) => new(new { Updated = status });
 
-    private IActionResult ReturnNotFound(string logMessage)
+    private IActionResult ReturnNotFoundWithLogging(string logMessage, LogLevel logLevel)
     {
-        _logger.LogWarning("{LogMessage}", logMessage);
+        _logger.Log(logLevel, "{LogMessage}", logMessage);
```
ashward, line 84: Not sure we need to pass in the log level, as it's only ever passed in as `LogLevel.Information`, so you could have just updated this line to
```suggestion
        _logger.LogInformation("{LogMessage}", logMessage);
```

ashward (approved):
Other than the suggestion about it being slightly over-engineered (which you can take or leave!), it looks okay to me.

UI/Pages/TuitionPartner.cshtml.cs
hasan3d: If we make the method more flexible, we could reuse it for any other log level. That was my thought process. Happy to apply your suggested approach, as well.

ashward: TBH this whole `ReturnNotFound` method is pretty superfluous anyway, I reckon!

hasan3d: Yeah, agree.

PR #231, opened 2023-01-30
NTP-838: Ensure that javascript is included in a consistent way

hasan3d:
## Context
A few of the newer javascript files have been added as page specific sub files and are not part of the webpack (so not minimised) or have any versioning (so are causing caching issues when updated).  

Some of the original javascript files were included in src folder and are in the webpack, so are minimised and versioned, but are not specific to the page.

- All javascript is included through the single webpack file
- Javascript is being conditionally applied to page elements using data-module
- We also use babel to make the JS more compatible
- Also carried out refactoring across the solution with the Clean Architecture in mind and moved the code to the appropriate layers.

## Guidance to review

Ensure that javascript is included in a consistent way. Also, ensure the changes proposed with this PR is satisfactory. Test with and without js enabled on mobile and check accessibility.

## Link to Jira ticket

[https://dfedigital.atlassian.net/browse/NTP-838](https://dfedigital.atlassian.net/browse/NTP-838)

## Things to check

- [x] Debug logging has been added after all logic decision points


UI/src/javascript/tuition-partner.js
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  const init = () => {
+    if (shortlistCheckboxes) {
+      shortlistCheckboxes
+        .querySelector("input[type='checkbox']")
+        .addEventListener("click", onCheckboxClick);
+    }
+  };
+  const onCheckboxClick = async (event) => {
```
ashward, line 29: Most of this logic is basically the same as `onCheckboxClick` for the search results. Could we combine these so we have a single place where the API is called?

hasan3d: Thanks for the suggestion. Sure, I will look into that.

UI/src/javascript/search-results.js
```diff
@@ -0,0 +1,126 @@
...
+    `/search-results?handler=RemoveShortlistedTuitionPartner&tuitionPartnerSeoUrl=${seoUrl}`;
+
+  const getSeoUrlFromCheckboxId = (id) => id && id.replace("shortlist-cb-", "");
+
+  const isFetchResponseDataValid = (response) =>
+    response &&
+    response.isCallSuccessful !== "undefined" &&
+    response.totalShortlistedTuitionPartners !== "undefined";
+
+  const shortlistCheckboxes = document.querySelectorAll(
```
ashward, line 19: If the shortlist checkboxes aren't in the DOM when this script is loaded and are later added, then they won't be included when `init()` is called. Given that `shortlistCheckboxes` isn't used outside of `init()` then you could just move this `document.querySelectorAll` into `init()`.

UI/src/javascript/search-results.js
```diff
@@ -0,0 +1,126 @@
...
+          badgeElement,
+          result.totalShortlistedTuitionPartners
+        );
+      }
+    } else {
+      checkBox.checked = elseCheckboxState;
+    }
+  };
+  const onChecked = async (checkbox) => {
+    return await axios
```
ashward, line 97: Is there a reason for switching from `fetch` to `axios`? Backward compatibility? It's not an issue (it's a perfectly reasonable change) but I just wondered what the driver was behind it.

UI/src/javascript/tuition-partner.js
```diff
@@ -0,0 +1,109 @@
...
+import axios from "axios";
+
+const TuitionPartner = (() => {
+  const getSeoUrlFromId = (id) => id && id.replace("shortlist-tpInfo-cb-", "");
+  const getAddToShortlistFetchUrl = (id) =>
+    `/tuition-partner/${id}?handler=AddToShortlist`;
+  const getRemoveFromShortlistFetchUrl = (id) =>
+    `/tuition-partner/${id}?handler=RemoveFromShortlist`;
+
+  const shortlistCheckboxes = document.querySelector(
```
ashward, line 12: Same thing as `search-results.js` about when the DOM is queried. Would be better to query it when `init()` is called rather than when the file is loaded.

ashward:
Thinking about this, it might be better to just have a single js file for the shortlisting feature, rather than a separate one per page?

UI/src/javascript/search-results.js
hasan3d: Yeah, it was mainly for backwards compatibility.

UI/src/javascript/api/search-results-client.js
```diff
@@ -0,0 +1,28 @@
+import BaseClient from "./base-client";
+import { getSeoUrlFromId } from "../util";
+
+const getAddShortlistTuitionPartnerUrl = (seoUrl) =>
+  `/search-results?handler=AddShortlistedTuitionPartner&tuitionPartnerSeoUrl=${seoUrl}`;
```
ashward, line 5: For safety it would be good to encode any values used directly in a URL like this using `encodeURIComponent()`.
e.g.:
```suggestion
  `/search-results?handler=AddShortlistedTuitionPartner&tuitionPartnerSeoUrl=${encodeURIComponent(seoUrl)}`;
```

UI/src/javascript/api/search-results-client.js
```diff
@@ -0,0 +1,28 @@
...
+
+const getRemoveShortlistedTuitionPartnerUrl = (seoUrl) =>
+  `/search-results?handler=RemoveShortlistedTuitionPartner&tuitionPartnerSeoUrl=${seoUrl}`;
```
ashward, line 8:
```suggestion
  `/search-results?handler=RemoveShortlistedTuitionPartner&tuitionPartnerSeoUrl=${encodeURIComponent(seoUrl)}`;
```

UI/src/javascript/api/tuition-partner-client.js
```diff
@@ -0,0 +1,34 @@
+import BaseClient from "./base-client";
+import { getSeoUrlFromId, addPostHeaderWithVerificationToken } from "../util";
+
+const replaceIdString = "shortlist-tpInfo-cb-";
+
+const postAddToShortlistUrl = (id) =>
+  `/tuition-partner/${id}?handler=AddToShortlist`;
```
ashward, line 7:
```suggestion
  `/tuition-partner/${encodeURIComponent(id)}?handler=AddToShortlist`;
```

UI/src/javascript/api/tuition-partner-client.js
```diff
@@ -0,0 +1,34 @@
...
+const postRemoveFromShortlistUrl = (id) =>
+  `/tuition-partner/${id}?handler=RemoveFromShortlist`;
```
ashward, line 9:
```suggestion
  `/tuition-partner/${encodeURIComponent(id)}?handler=RemoveFromShortlist`;
```

UI/src/javascript/tuition-partner.js
hasan3d: done

Application/Common/Models/AddToShortlistModel.cs
```diff
@@ -0,0 +1,6 @@
+namespace Application.Common.Models;
+
+public class AddToShortlistModel : ShortlistModelCommon
```
ashward, line 3: Are these models only used for the tests? If so it might be worth moving them to the Tests namespace?

Application/Handlers/GetTuitionPartnerQueryHandler.cs
```diff
@@ -29,7 +40,7 @@ public class GetTuitionPartnerQueryHandler : IRequestHandler<GetTuitionPartnerQu
         var tp = tpResult.Result.Data.FirstResult;
 
         var subjects = tp.SubjectsCoverage!.Select(x => x.Subject).Distinct().GroupBy(x => x.KeyStageId)
-            .Select(x => $"{((Enums.KeyStage)x.Key).DisplayName()} - {x.DisplayList()}");
+            .Select(x => $"{EnumExtensions.DisplayName(((Enums.KeyStage)x.Key))} - {x.DisplayList()}");
```
ashward, line 43: Is there a reason behind not using the `DisplayName` as an extension method here?

Application/Common/Models/RemoveFromShortlistModel.cs
```diff
@@ -0,0 +1,6 @@
+namespace Application.Common.Models;
+
+public class RemoveFromShortlistModel : ShortlistModelCommon
```
ashward, line 3: Are these models only used for the tests? If so it might be worth moving them to the Tests namespace?

Application/Common/Models/AddToShortlistModel.cs
hasan3d: The `AddToShortlistModel` and `RemoveFromShortlistModel` models are used to capture the data from the client side. For example, for the following `Task<IActionResult> OnPostAddToShortlist([FromBody] AddToShortlistModel shortlistModel)`, when the user adds a TP to the shortlist, we will be receiving post body data with the `AddToShortlistModel` model, and we then process that on the server side.

Application/Common/Models/RemoveFromShortlistModel.cs
hasan3d: Please see my response above.

Application/Common/Models/ShortlistModelCommon.cs
```diff
@@ -0,0 +1,8 @@
+namespace Application.Common.Models;
+
+public class ShortlistModelCommon
```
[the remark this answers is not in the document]
hasan3d: Please see my response above.

Application/Handlers/GetTuitionPartnerQueryHandler.cs
hasan3d: We should be able to use the `DisplayName` as an extension method just fine. I will correct that now. Thanks!

Application/Application.csproj
```diff
@@ -16,6 +16,7 @@
 
     <ItemGroup>
         <PackageReference Include="MediatR.Extensions.Microsoft.DependencyInjection" Version="10.0.1" />
+        <PackageReference Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.2.0" />
```
kieranohanlon, line 19: Is this added for the "FromQuery" attribute?  Just wondering if there's a way we can do it without adding aspnet dll's in to the application layer?

UI/src/javascript/shortlist-checkbox.js
```diff
@@ -0,0 +1,112 @@
+"use strict";
+
+import ShortlistClient from "./api/shortlist-client";
+import { throwError } from "./util";
```
kieranohanlon, line 4: Not needed now it's in the base?

hasan3d: Good spot. I made a commit to remove the unused import.

Application/Application.csproj
hasan3d: done!

kieranohanlon (approved):
lgtm

PR #234, opened 2023-02-02
NTP-869: Shortlist: Carry search subject filter through to the shortlist and add group size column

kieranohanlon:
## Context

Shortlist: Carry search subject filter through to the shortlist and add group size column

Carry through the key stage and subject search filters and apply these to the shortlist data.

The existing key stage and subject columns are to be removed, since these will always contain the same data now the filter is applied.

The details of the searched key stage and subjects are added between the level 1 heading and the refine header.  A specific message (to be supplied by content) is shown if there are no filter values passed through.

The subject filtering is implemented using “AND” logic – a shortlisted TP needs to have all the subjects from the search to have data/prices returned.  

Meaning that returning to the search results, changing the filter and returning to the shortlist may mean that the TP will have no data applicable for the filtered subjects. These will be shown at the bottom of the table, in the same way as when the TP has no data for the group size or tuition type refinement filters.

Add the group size column in to the table, since it is a refinement option.

## Guidance to review

AC1: The search filter selections are brought in and applied to the shortlist page

GIVEN a search has been made, search subjects have been filtered upon and some TPs have been added to the shortlist

WHEN the shortlist is displayed

THEN the searched criteria is displayed on the shortlist page (as per design and content) and the data is filtered by these subjects

AC2: The shortlist still works if there are no subjects filtered on the search page

GIVEN a search has been made, the search subjects filters have been removed and some TPs have been added to the shortlist

WHEN the shortlist is displayed

THEN the searched criteria is not displayed on the shortlist page and the data is no longer filtered for the subjects selected, the data is for any key stage subject that the TPs offer.  We show a message specificity for this scenario (as supplied by the content team).

AC3: The shortlist uses “AND” logic when applying the key stage subject filter

GIVEN a search has been made, a single search subject has been filtered upon and then lots of TPs have been added to the shortlist.  Then the search subject filter is changed so that lots of subjects are filtered upon and the TPs search list is updated (to exclude some previously shortlisted TPs)

WHEN the shortlist is displayed

THEN the table will include all of the shortlisted TPs.  Any that do not offer all the filtered subjects will be displayed at the bottom of the shortlist table, in the order that they were added to the shortlist, with no data and an appropriate message (supplied by content team).

AC4: The previous key stage and subject columns are no longer shown

GIVEN a search has been made, search subjects have been filtered upon and some TPs have been added to the shortlist

WHEN the shortlist is displayed

THEN the displayed table no longer shows the key stag e column or the subjects column

AC5: The group sizes are now added as a new column

GIVEN a search has been made, search subjects have been filtered upon and some TPs have been added to the shortlist

WHEN the shortlist is displayed

THEN the table now includes the group size column (as per design and content) and the data in this column is updated as the refinement filters are applied.

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-869

## Things to check


Application/Extensions/StringExtensions.cs
```diff
@@ -46,4 +46,16 @@ private static string RegexReplace(this string value, string pattern, string rep
             value, pattern, replacement,
             RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
 
+    public static string DisplayList(this IEnumerable<string>? strings)
```
hasan3d, line 49: Can we add a unit test for this method?

Application/Extensions/KeyStageExtensions.cs
```diff
@@ -6,12 +6,7 @@ public static class KeyStageExtensions
...
     public static string DisplayList(this IEnumerable<KeyStage> keyStages)
     {
-        var commaSeparated = string.Join(", ", keyStages.Select(e => e.Id).Distinct().OrderBy(e => e));
-
-        var lastCommaIndex = commaSeparated.LastIndexOf(",", StringComparison.Ordinal);
-
-        if (lastCommaIndex == -1) return commaSeparated;
-
-        return commaSeparated.Remove(lastCommaIndex, 1).Insert(lastCommaIndex, " and");
+        var keyStageStrings = keyStages.Select(e => e.Id.ToString()).Distinct().OrderBy(e => e);
```
hasan3d, line 9: Should we consider better naming for this method?

Infrastructure/LookupDataService.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Application;
```
hasan3d, line 1: Do we need this service? I suggest we create a Mediator query, i.e. `GetSubjectsAsyncQuery` with a handler to get the data and inject the `ILookupDataRepository` to fetch the data from the repository. Then we can use `await _mediator.Send(query);` to get the data we need.

Application/Extensions/KeyStageExtensions.cs
kieranohanlon: Not sure this is really worth picking up, but have removed the variable anyway.

Infrastructure/LookupDataService.cs
kieranohanlon: This has been done to be consistent with the other services (e.g. TuitionPartnerService and LookupDataService).  I think that removing the services and going straight to repositories is something we should consider, but is a wider refactor job to ensure it is consistent.  
I think that the calls to the service are already done async via mediator, but to the service.
I did notice that this file was not in the same folder as the TuitionPartnerService and LookupDataService classes, so have moved it there.

hasan3d: Okay, I suggested it because we are adding a new service. We may have to refactor these at some point (i.e., TuitionPartnerService and LookupDataService) and use the Mediator query or command. I am not suggesting refactoring the existing other services but not adding new ones.

kieranohanlon: That lookup repository wasn't being called prior to this, but was in existence.  So to use it I either 

a) stick with the existing pattern of going via a service and accept that I'm adding it, even if it might be removed later
b) just call the repository directly and don't refactor which will be a different pattern to the existing
c) refactor all to use the approach we want to use.  

I'd prefer to stick with a) for now and move to c) when we have a chance.

Application/ILookupDataService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Domain;
+using Domain.Search;
+
+namespace Application;
+
+public interface ILookupDataService
```
hasan3d, line 6: Can we move this interface under the `Common/Interfaces` folder? Thanks!

kieranohanlon: Done

hasan3d (approved):
lgtm

PR #308, opened 2023-03-28
NTP-1038: Contact us (and other) pages allow random URLs into the 'Back' link which could be open to abuse

kieranohanlon:
## Context

Contact us (and other) pages allow random URLs into the 'Back' link which could be open to abuse

Several pages within the site set the URL for the ‘Back’ link as the Referer header.

This could potentially be open to abuse - if someone could manipulate the Referer to point to a malicious site.

The pages are:

Contact us

Cookies

Funding/reporting

possibly others…

Suggested solution

Update the following “support pages”

Contact us

Accessibility

Cookies

Privacy

Report Issues

Funding and reporting

Rather than the referer for the back link pass in the return URL as part of the query string

If there is no return url then show the Home link

If the user skips from support page to support page we do not build up the return URL, the back link will return to the last non-support page

It was also noticed that the cookie page cookie banner could be updated to use the from URL and that the cookie page could be - when a user sets their cookie preference from within the cookie page the link at the top of the page should return them to their previous page in the same way as the back button.

An advantage of using the return URL is that TempData is no longer needed, so the cookie page has been updated to remove this cookie from the content of the page.

Work done as part of https://github.com/DFE-Digital/find-a-tuition-partner/pull/308

## Guidance to review

Ensure that the back links on the pages above still work from various pages and that the cookie preferences work

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1038

## Things to check

### DIFF
--- a/.github/workflows/deploy-to-gpaas.yml
+++ b/.github/workflows/deploy-to-gpaas.yml
@@ -75,7 +75,7 @@ jobs:
 
       - name: Install the CF CLI
         run: |
-          wget -v -O cf.tar.gz "https://s177d01corec7afkt3qxq7ie.blob.core.windows.net/ash-temp-cloudfoundry-cli/cf8-cli_8.6.1_linux_x86-64.tgz"
+          wget -v -O cf.tar.gz "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=v8"
           sudo tar xzf cf.tar.gz --wildcards --directory /usr/local/bin/ "cf*"
 
       - name: Authenticate
--- a/.github/workflows/import-data.yml
+++ b/.github/workflows/import-data.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
       - name: Install the CF CLI
         run: |
-          wget -v -O cf.tar.gz "https://s177d01corec7afkt3qxq7ie.blob.core.windows.net/ash-temp-cloudfoundry-cli/cf8-cli_8.6.1_linux_x86-64.tgz"
+          wget -v -O cf.tar.gz "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=v8"
           sudo tar xzf cf.tar.gz --wildcards --directory /usr/local/bin/ "cf*"
 
       - name: Authenticate
--- a/.github/workflows/pull-request-teardown.yml
+++ b/.github/workflows/pull-request-teardown.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - name: Install the CF CLI
       run: |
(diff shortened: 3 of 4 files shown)


UI/Pages/Enquiry/Build/CheckYourAnswers.cshtml.cs
```diff
@@ -95,8 +95,8 @@ public async Task<IActionResult> OnPostAsync()
 
         if (!submittedConfirmationModel.IsValid && submittedConfirmationModel.ErrorStatus == HttpStatusCode.InternalServerError.ToString())
         {
-            TempData["Status"] = HttpStatusCode.InternalServerError;
-            return RedirectToPage(nameof(ErrorModel));
+            //This is a temporary fix, until errors are addressed in NTP-1044
+            throw new InvalidDataException("Error on check you answers");
```
hasan3d, line 99: We can create an `InternalServerErrorResult` class and redirect the user to the Error page as we do for the 404 `NotFound` stuff.. i.e: 
```
/// <summary>
/// Represents an <see cref="StatusCodeResult"/> that when
/// executed will produce a Not Found (500) response.
/// </summary>
/// 
[DefaultStatusCode(DefaultStatusCode)]
public class InternalServerErrorResult : StatusCodeResult
{
    private const int DefaultStatusCode = StatusCodes.Status500InternalServerError;

    /// <summary>
    /// Creates a new <see cref="NotFoundResult"/> instance.
    /// </summary>
    public InternalServerErrorResult() : base(DefaultStatusCode)
    {
    }
}
```
We can just call `return new InternalServerErrorResult(),` which will redirect the user to the Error page with a 500 status code.

kieranohanlon: This is a temporary fix, until errors are addressed in NTP-1044.  With my intention being that unexpected exceptions will be allowed to propagate.

hasan3d (dismissed):
lgtm

ashward (commented):
There's a potential issue in that the links in the query string are `http` rather than `https`. It also allows random schemes (e.g. https://find-a-tuition-partner-pr-308.london.cloudapps.digital/contact-us?FromReturnUrl=ftp://find-a-tuition-partner-pr-308.london.cloudapps.digital/ ). Not sure if this is particularly exploitable, but it feels wrong, and maybe we should always have them as URLs relative to the site root rather than including the scheme and domain? This would also mitigate the issue of having to validate the scheme and domain.

kieranohanlon:
@ashward  - I have updated the code, so that it ensures that it's always https.  
Locally it was using https as the return URL, but not when deployed for some reason.  As you prob know, the site does do a redirect to https if someone hits http, but ideally we'd not rely on this redirect when it's not needed.
We originally had the relative path, the reason that we now include the full URL (inc. host etc) in the return is so that we can ensure that the full URL is from the NTP site.

PR #327, opened 2023-04-24
NTP-1136: Add HTTP Basic authentication to the lower environments

hasan3d:
## Context
At the moment all our lower environments (PR, QA, Research and Staging) are completely unauthenticated.

This didn’t matter too much when it was a read-only service, but as it’s now transactional it would be good to put some control in place.

Bind the QA route service to the application route for the PR environment. With these changes in place, we will now ask for basic HTTP authentication with login and password.

See here: https://find-a-tuition-partner-pr-327.london.cloudapps.digital/

## Guidance to review

Navigate to the following link: https://find-a-tuition-partner-pr-327.london.cloudapps.digital. It should ask for basic HTTP authentication, and If you then use the correct credential, you should be able to use the service as usual.

## Link to Jira ticket

[https://dfedigital.atlassian.net/browse/NTP-1136](https://dfedigital.atlassian.net/browse/NTP-1136)

## Things to check


.github/workflows/pull-request.yml
```diff
@@ -78,6 +78,10 @@ jobs:
           --var google-drive-shared-drive-id="${{ secrets.GOOGLE_DRIVE_SHARED_DRIVE_ID }}" \
           --var google-drive-tuition-partner-data-folder-id="${{ secrets.GOOGLE_DRIVE_TUITION_PARTNER_DATA_FOLDER_ID }}" \
           --var google-drive-tuition-partner-logo-folder-id="${{ secrets.GOOGLE_DRIVE_TUITION_PARTNER_LOGO_FOLDER_ID }}"
+        
+    - name: Bind route service to the application route
+      run: |
+        cf bind-route-service london.cloudapps.digital find-a-tuition-partner-qa-basic-auth --hostname find-a-tuition-partner-pr-${{ github.event.number }}
```
ashward, line 84: Don't we need to have unique names for the route service?
```suggestion
        cf bind-route-service london.cloudapps.digital find-a-tuition-partner-pr-${{ github.event.number }}-basic-auth --hostname find-a-tuition-partner-pr-${{ github.event.number }}
```

.github/workflows/pull-request-teardown.yml
```diff
@@ -33,6 +33,10 @@ jobs:
         cf auth
         cf target -o "${CF_ORG}" -s "${CF_SPACE}"
 
+    - name: Un Bind route service to the application route
+      run: |
+        cf unbind-route-service london.cloudapps.digital find-a-tuition-partner-qa-basic-auth --hostname find-a-tuition-partner-pr-${{ github.event.number }}
```
ashward, line 38:
```suggestion
        cf unbind-route-service london.cloudapps.digital find-a-tuition-partner-pr-${{ github.event.number }}-basic-auth --hostname find-a-tuition-partner-pr-${{ github.event.number }}
```

UI/cypress/support/commands.js
```diff
@@ -24,19 +24,30 @@
...
+    });
 
-  if (username && password) {
-    options = options || {};
-    options.auth = {
-      username: username,
-      password: password,
+    options = {
+      ...options,
+      headers: {
```
ashward, line 44: Not sure why you're managing the headers yourself. Cypress already handles this for you if you set the `auth` option. https://docs.cypress.io/api/commands/visit#Add-basic-auth-headers

.github/workflows/pull-request.yml
hasan3d: In this case, we don't, as we are using the same backing service, `find-a-tuition-partner-qa-basic-auth` for the qa and pr, which we provisioned via the command line.

.github/workflows/pull-request-teardown.yml
hasan3d: In this case, we don't, as we are using the same backing service, `find-a-tuition-partner-qa-basic-auth` for the qa and pr, which we provisioned via the command line.

.github/workflows/end-to-end-testing.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         env:
           CYPRESS_username: ${{ inputs.username }}
```
ashward, line 40: We should get rid of these as they will no longer work

.github/workflows/pull-request.yml
ashward: Ah okay - so the one route service can handle the routes for multiple applications. Cool.

UI/cypress/support/commands.js
```diff
@@ -24,25 +24,18 @@
...
-      username: username,
-      password: password,
-    };
-  }
-
+  options = applyBasicAuth(options);
   return originalFn(url, options);
 });
 
+Cypress.Commands.add("requestWithBasicAuth", (options) => {
```
ashward, line 34: Could we not just use `Cypress.Commands.overwrite("request"...` much like we do with visit? That would make it work with any `request` in the future.

ashward, line 34:
```suggestion
Cypress.Commands.overwrite("request", (originalFn, options) => {
```

UI/cypress/support/commands.js
```diff
@@ -24,25 +24,18 @@
...
+  options = applyBasicAuth(options);
+  return cy.request(options);
```
ashward, line 36:
```suggestion
  return originalFn(options);
```

UI/cypress/support/commands.js
hasan3d: I tried that [approach](https://github.com/DFE-Digital/find-a-tuition-partner/pull/327/commits/cb9f7444385014e0b8b8fd4d762a25bb13618301); however, It didn't work. Looking at the [logs](https://github.com/DFE-Digital/find-a-tuition-partner/actions/runs/4813444926/jobs/8570013729), some of our end-to-end tests make a call to the external URL, and with our basic auth in the header in the request they failed gives us 403 error.

hasan3d: please see my comment above

UI/cypress/support/commands.js
hasan3d: please see my comment above

UI/cypress/support/commands.js
ashward: I see what you mean - `request` has a much more complicated signature which would need to be replicated, and is also used to access external resources. Maybe try something like this?
```js
Cypress.Commands.overwrite("request", (originalFn, ...args) => {
  if(typeof args[0] == 'object') {
    let options = {...args[0]};

    const url = new URL(options.url, Cypress.config().baseUrl);

    if(url.toString().startsWith(Cypress.config().baseUrl)) {
      options = applyBasicAuth(options);
    }

    return originalFn(options);
  }

  return originalFn.apply(null, args);
});
```

UI/cypress/support/commands.js
```diff
@@ -24,23 +24,23 @@
...
+});
 
-  if (username && password) {
-    options = options || {};
-    options.auth = {
-      username: username,
-      password: password,
-    };
+Cypress.Commands.overwrite("request", (originalFn, options) => {
+  if (Cypress.config().baseUrl.includes("find-a-tuition-partner")) {
```
ashward, line 40: This seems wrong - the baseUrl is always going to be the same regardless of whether the url we're requesting is our site or not isn't it? What we need to be checking is whether the URL we're requesting (which may or may not be an external site) is our site, and only sending the credentials if it is ours. Otherwise we risk sending our credentials to random other sites.

ashward: This is what I was trying to do with my bit of code here: https://github.com/DFE-Digital/find-a-tuition-partner/pull/327#discussion_r1179633489

hasan3d: Okay, I have now incorporated your [suggested changes](https://github.com/DFE-Digital/find-a-tuition-partner/pull/327#discussion_r1179633489) to this PR. Made a slight tweak on the below:
```
if(typeof args[0] == 'object') to
if (typeof args == "object") 
```
Cypress will enforce basic authentication on all requests to the base URL after overwriting the "request" command with these modifications.

PR #332, opened 2023-05-03
NTP-1185: Update the tribal spreadsheet importer to use a single spreadsheet rather than one per TP

kieranohanlon:
## Context

Update the Tribal spreadsheet importer to use a single spreadsheet rather than one per TP

The existing TP spreadsheet importer assumes that each TP is in their own spreadsheet.

There is a request from Tribal that we merge all the TP data in to one spreadsheet.  Although the TP logos will still remain as separate files.

The following approach can be applied to the code:

- The existing code to open the spreadsheet and the code to validate and process the spreadsheet can remain almost as is.
- The existing TribalSpreadsheetTuitionPartnerFactory.GetTuitionPartner method could be changed so that it returns a collection of TPs rather than a single one - so the whole spreadsheet would be processed with this single call.
- The existing DataImporterService that loops through the enumerated data files could be updated to ensure that it just processes the single file.  The IDataFileEnumerable could be removed, although possibly consider leaving and using this for now, should we revert to to multiple files in the future?
- Consider removing the SpreadsheetTuitionPartnerFactory and QualityAssuredSpreadsheetTuitionPartnerFactory, since these are no longer needed (they are the old spreadsheet processors).
- Ensure that the error handing is in place if there is no spreadsheet to process and that it will throw an error and rollback.
- Ensure that the existing matching on the Tribal ID is case insensitive
- Consider doing some performance comparisons between the existing and updated importer to ensure that the changes do not slow things down too much.

## Guidance to review

Test the importer with the new spreadsheet supplied by Tribal

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1185

## Things to check


hasan3d (approved):
lgtm

PR #336, opened 2023-05-05
NTP-1091: Layout Update | PII warning message

kieranohanlon:
## Context

Move the PII to below the heading

Move the PII to below the heading

## Guidance to review

Give these pages a good test.  Since the aspnet component (https://github.com/gunndabad/govuk-frontend-aspnetcore/blob/main/docs/components/character-count.md) did not allow the PII warning to be placed in that desired location.  So the cshtml had to be changed to use the GDS markup (https://design-system.service.gov.uk/components/character-count/) to match the previous functionality, but allow the PII warning to move.

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1091

## Things to check


UI/src/sass/enquiry-build.scss
```diff
@@ -11,3 +11,8 @@
     width: 25%;
   }
 }
+
+.pii-warning {
+  margin-bottom: 10px;
```
hasan3d, line 16: Using "em" units will make the margin adjust proportionally based on the font size of the element or the root element, respectively. This ensures that the margin will be consistent across different devices and screen sizes.

For example:
```
.pii-warning {
  margin-bottom: 0.53em;
  margin-top: 0.53em;
}
```

kieranohanlon: I have updated this as suggested, although I don't really think it was worth raising or doing since I think that px is used a lot more than em in the existing .scss.

PR #337, opened 2023-05-10
NTP-1205: Prevent multiple enquiries from submitting on double click with js disabled

kieranohanlon:
## Context

Prevent multiple enquiries from being submitting on double click when js is disabled

We have now had 2 sets of enquires where multiple versions of the same enquiry have been created.

The only way that we have identified that this could happen is by having javascript disabled and multiple clicking (within a few seconds, before response is returned) on the enquiry check your answers page.

We need to prevent this from happening (possibly locking the enquiry when started processing to prevent additional requests) server side.

A further consideration is that the final click should ideally return the processed enquiry (from the initial click) details, so that the user can see the submitted confirmation page with the enquiry ref number as normal.

## Guidance to review

Ensure that the normal functionality works as expected.  Then test this with multiple clicks when js is disabled and ensure that only 1 enquiry is submitted.  The same multiple submission check has been applied to the TP response POST, so this should also be tested.

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1205

## Things to check


UI/Services/DistributedSessionService.cs
```diff
@@ -1,12 +1,16 @@
...
 using Newtonsoft.Json;
+using Polly;
+using IntegerConstants = UI.Constants.IntegerConstants;
 
 namespace UI.Services;
 
 public class DistributedSessionService : ISessionService
 {
     private const string DefaultPreKey = "General";
+    public const string FormPostPreKey = "FormPostPreKey";
```
hasan3d, line 12: make it `public` as the other one

UI/Services/DistributedSessionService.cs
```diff
@@ -121,7 +125,8 @@ public async Task<bool> SessionDataExistsAsync(string preKey = DefaultPreKey)
     public async Task<bool> AnySessionDataExistsAsync()
     {
         await LoadDataFromDistributedDataStore();
-        return _contextAccessor!.HttpContext!.Session!.Keys.Any();
+        var sessionKey = GetSessionKey(FormPostPreKey);
+        return _contextAccessor!.HttpContext!.Session!.Keys.Where(x => x != sessionKey).Any();
```
hasan3d, line 129: We can rewrite this with the below:

`return _contextAccessor!.HttpContext!.Session!.Keys.Any(x => x != sessionKey);`

Application/Common/Interfaces/ISessionService.cs
```diff
@@ -11,4 +11,10 @@ public interface ISessionService
     Task<bool> SessionDataExistsAsync(string preKey = DefaultPreKey);
     Task<bool> AnySessionDataExistsAsync();
     Task ClearAllAsync();
+    Task<bool> IsDuplicateFormPostAsync(string formPostTimestampKey = "FormPostTimestamp");
+    Task StartFormPostProcessingAsync(string formPostTimestampKey = "FormPostTimestamp");
+    Task SetFormPostResponseAsync<T>(T postResponseModel, string formPostModelKey = "FormPostModelKey");
+    Task<T> GetPreviousFormPostResponseAsync<T>(string formPostModelKey = "FormPostModelKey");
+    Task SetAsync<T>(string key, T value, string preKey = DefaultPreKey);
```
hasan3d, line 18: If we're not using it in any context other than DistributedSessionService, it seems unnecessary.

Application/Common/Interfaces/ISessionService.cs
```diff
@@ -11,4 +11,10 @@ public interface ISessionService
...
+    Task<T?> GetAsync<T>(string key, string preKey = DefaultPreKey);
```
hasan3d, line 19: If we're not using it in any context other than DistributedSessionService, it seems unnecessary.

UI/Services/DistributedSessionService.cs
```diff
@@ -131,6 +136,84 @@ public async Task ClearAllAsync()
...
+
+    public async Task SetFormPostResponseAsync<T>(T postResponseModel, string formPostModelKey = "FormPostModelKey")
+    {
+        await SetAsync(formPostModelKey, postResponseModel, FormPostPreKey);
+    }
+
+    public async Task<T> GetPreviousFormPostResponseAsync<T>(string formPostModelKey = "FormPostModelKey")
+    {
+        //Use Polly to retry 
+        int numberOfRetries = IntegerConstants.SecondsClassifyAsDuplicateFormPost;
```
hasan3d, line 171: To avoid confusion, we should not reuse the constant value of `SecondsClassifyAsDuplicateFormPost` as the value for the variable name maxNumberOfRetries. Instead, we should choose a different value, such as `5`, to explicitly specify the maximum number of retries. So, the variable name should be maxNumberOfRetries and we can set its value as `5` if we want.

UI/Pages/Enquiry/Respond/CheckYourAnswers.cshtml.cs
```diff
@@ -61,24 +63,45 @@ public async Task<IActionResult> OnPostAsync()
...
+        var enquiryResponseConfirmationModelKey = $"{EnquiryResponseConfirmationModelKey}-{Data.SupportReferenceNumber}";
+        var enquiryResponseFormPostTimestampKey = $"{EnquiryResponseFormPostTimestampKey}-{Data.SupportReferenceNumber}";
+
+        var isDuplicateFormPost = await _sessionService.IsDuplicateFormPostAsync(enquiryResponseFormPostTimestampKey);
+
+        if (!isDuplicateFormPost)
         {
-            return NotFound();
-        }
+            if (!isValidMagicLink)
```
hasan3d, line 75: To avoid unnecessary calls to isDuplicateFormPost, we ought to extract the validation logic for the magic link from !isDuplicateFormPost. This would enable us to return "not found" immediately when an invalid magic link is encountered.

UI/Services/DistributedSessionService.cs
kieranohanlon: Assuming mean `private`?  I have done that.

UI/Services/DistributedSessionService.cs
kieranohanlon: Done

Application/Common/Interfaces/ISessionService.cs
kieranohanlon: I think that these may be used in the future, but agree, if not being used for now they can be removed from interface - so I have done that.

Application/Common/Interfaces/ISessionService.cs
kieranohanlon: As above - done

UI/Services/DistributedSessionService.cs
kieranohanlon: I wasn't to sure what you were suggesting here.  Since we retry every second I think it should use the same int constant that we classify as a duplicate form post, otherwise there's a danger a different constants could be out of sync.  We should retry for as long as we classify the duplicate post is the intention.  I have updated the variable name to be more meaningful, to see if that helps.

hasan3d (approved):
lgtm!

hasan3d (approved):
lgtm

PR #350, opened 2023-06-08
(New)NTP-1151: Delay sending the TP emails until the enquirer email has been successfully delivered

kieranohanlon:
## Context

Delay sending the TP emails until the enquirer email has been successfully delivered?

Delay sending the TP emails until the enquirer email has been successfully delivered, but also ensure that certain failed emails are retried.  This has involved significant changes to the existing code, to refactor the email sending so that it logs emails in the database and then sends the email separately.

<html>
<body>
<p data-pm-slice="1 3 []">The following approach has been discussed with the dev team and will be implemented…</p><p></p><p>Decouple the sending of emails in to 2 steps:</p><ol class="ak-ol"><li><p>Log the emails to be sent - this will be a log that has all the email details (e.g. email address, Notify template, Notify parameters/personalisation)</p></li><li><p>Send the emails and manage the responses from Notify</p></li></ol><p></p><p>The advantages of decoupling this process means that we can add emails to the log to be sent (step 1 above), but can then decide if we want to send the emails (step 2 above) immediately, dependant upon if another email has been delivered or even schedule an email to be sent at a future date (e.g. batch up the TP response emails that go to the enquirers).  Step 2 will also allow us to retry any failed emails that should be resent.</p><p>The emails will be added to the log in step 1 as/when needed, so this will be at the same point as the current emails are being sent within the code.  The slight difference to the existing logic will be that we will ensure that the TP emails for a new enquiry that are logged will not be sent immediately.  These will only be sent once a successfully delivered enquirer email has been recorded in the logs.</p><p>Step 2 above can be called immediately within the code, after step 1 when needed.  So this means that after the enquirer email has been logged we will call step 2 to send it immediately, as currently happens, ensuring that the enquirer email is sent straight away.</p><p>But step 2 will also be called via a polling job (see details below), this allows us to get the latest status details for any outstanding emails and also send any that are ready to be processed.  Such as sending TP emails once the enquirer email has been delivered, sending the delayed batch emails or even retrying certain failed emails.</p><p></p><p>Polling to update email status and send any that require processing:</p><ul class="ak-ul"><li><p>This will be a simple webpage GET request that will do the processing.  It will be setup to be called (possibly via a CURL job) on a regular basis (e.g. every minute).  Should a user somehow manage to find out that this page exists it should not do any harm if they also call the page.  This page may also be useful to request manually during testing.</p></li><li><p>The GetNotification methods on the Notify client will allow us to get the email status from Notify - see <a data-inline-card="" href="https://github.com/alphagov/notifications-net-client/blob/main/src/GovukNotify/Interfaces/IAsyncNotificationClient.cs" data-card-data="">https://github.com/alphagov/notifications-net-client/blob/main/src/GovukNotify/Interfaces/IAsyncNotificationClient.cs</a> </p></li><li><p>Notify does have a Callback feature (see <a data-inline-card="" href="https://docs.notifications.service.gov.uk/net.html#callbacks" data-card-data="">https://docs.notifications.service.gov.uk/net.html#callbacks</a> ), which could POST an update to our website, rather than needing to poll on a regular basis.  The difficulty being that there is only one Notify environment and multiple NTP environments (dev, QA, staging, prod etc), so the Notify Callback wouldn’t be able to update the various NTP environment logs.</p></li><li><p>Rather than a GET page for polling in the future it may be desirable to use a microservice architecture with an Azure Logic App and recurrence trigger, and possibly use a service bus.  This may be an approach that we can move to when the service is fully migrated to Azure, this should be considered when architecting the code to ensure it is segregated and portable if needed.</p></li></ul><p></p><p>To ensure that we have a unique Notify reference for each email we’ll prefix the exiting Notify reference with the non-production website environment name.  This has the added advantage that it will make the source of emails even more obvious within Notify.  This is following on from <a data-inline-card="" href="https://dfedigital.atlassian.net/browse/NTP-1152" data-card-data="">https://dfedigital.atlassian.net/browse/NTP-1152</a>  - so references will now be along the lines of <span class="code" spellcheck="false">&lt;env-name&gt;-&lt;refnum&gt;-erne-&lt;tpurl&gt;</span> for non production and <span class="code" spellcheck="false">&lt;refnum&gt;-erne-&lt;tpurl&gt;</span> for production.</p><p></p><p>To ensure that the polling does not run at the same time as the previous polling we considered locking the records.  But for now a simple “ScheduledProcessingInfo“ table has been added that records when  the email processing starts and finishes and does not start a new one until the last one has finished.  If the previous one has hit issues the polling finish date has not been updated then the subsequent polls will log a warning (after 5 minutes) then log an error (after 30 minutes) and will start processing again after this.</p><p></p><p>Within the code, we’ll use a couple of config values to decide if want to have the dependency and delayed/batching active or not.  This will allow us to remove the delay from certain environments (e.g. local dev) and have email sent immediately, as is currently the case.</p><p>We will also add a new config entry to switch off the email sending completely, which will be useful for testing when the links required are shown on the confirmation pages, so the emails are not required unless they are being tested specifically.</p><p></p><p>One difficultly is catering for the historical/existing email data, since all enquiries and responses will have a non-nullable FK link to the email log (see code below).  For historical/existing emails we will add a dummy EmailLog record (see example of dummy below) and link them to this.</p><p></p><p>Domain object implementation:</p><p>ScheduledProcessingInfo is used to ensure that the previous polling process has finished running before the next one starts</p><pre><code>public class ScheduledProcessingInfo
{
    public int Id { get; set; }
    public string ScheduleName { get; set; } = null!;
    public DateTime LastStartedDate { get; set; } = DateTime.UtcNow;
    public DateTime? LastFinishedDate { get; set; }
    public string? Status { get; set; }
}</code></pre><p>EmailLog class, that will drive the emails that get sent.</p><pre><code>    public class EmailLog
    {
        public int Id { get; set; }
        public DateTime CreatedDate { get; set; } = DateTime.UtcNow;
        public DateTime? ProcessFromDate { get; set; } //Used for batching, will only send the email after this date.
                                                       //If null then the email is not sent (e.g. awaiting response from enquirer before sending to TPs)
        public DateTime? LastEmailSendAttemptDate { get; set; } //This will be when NTP 1st sends the email.
                                                                //It will be updated with any retries if there have been issues.
        public DateTime FinishProcessingDate { get; set; } = DateTime.UtcNow.AddDays(2); //This is how long we want to try and send emails for
        public DateTime? LastStatusChangedDate { get; set; } //If status changes we update this

        public string EmailAddress { get; set; } = null!;
        public string EmailTemplateShortName { get; set; } = null!;
        public string ClientReferenceNumber { get; set; } = null!;
        public string? EmailAddressUsedForTesting { get; set; }

        public int EmailStatusId { get; set; }

        public EmailStatus EmailStatus { get; set; } = null!;
        public EmailNotifyResponseLog? EmailNotifyResponseLog { get; set; }
        public ICollection&lt;EmailPersonalisationLog&gt;? EmailPersonalisationLogs { get; set; }
        public ICollection&lt;EmailLogHistory&gt;? EmailLogHistories { get; set; }

        public Enquiry? EnquirerEnquirySubmitted { get; set; }
        public ICollection&lt;TuitionPartnerEnquiry&gt;? TuitionPartnerEnquiriesSubmitted { get; set; }
        public EnquiryResponse? EnquirerResponse { get; set; }
        public EnquiryResponse? TuitionPartnerResponse { get; set; }

        public EmailTriggerActivation? ThisEmailActivationTriggeredBy { get; set; }
        public ICollection&lt;EmailTriggerActivation&gt;? EmailsActivatedByThisEmail { get; set; }
    }</code></pre><p></p><p>EmailPersonalisationLog - this contains the Notify personalisation data, to be sent with the emails when they are processed</p><pre><code>    public class EmailPersonalisationLog
    {
        public int Id { get; set; }
        public string Key { get; set; } = null!;
        public string Value { get; set; } = null!;

        public int EmailLogId { get; set; }
        public EmailLog EmailLog { get; set; } = null!;
    }</code></pre><p></p><p>EmailLogHistory - this is mainly used to store a log of when emails have been sent or their status has changed</p><pre><code>    public class EmailLogHistory
    {
        public int Id { get; set; }
        public DateTime CreatedAt { get; set; } = DateTime.UtcNow;

        public DateTime? ProcessFromDate { get; set; }
        public DateTime? LastEmailSendAttemptDate { get; set; }

        public int EmailStatusId { get; set; }

        public int EmailLogId { get; set; }
        public EmailLog EmailLog { get; set; } = null!;
    }</code></pre><p></p><p>EmailNotifyResponseLog - this logs the success and exceptions that are returned upon the immediate sending of the email via Notify (see <a data-inline-card="" href="https://docs.notifications.service.gov.uk/net.html#send-an-email-response" data-card-data="">https://docs.notifications.service.gov.uk/net.html#send-an-email-response</a> )</p><pre><code>    public class EmailNotifyResponseLog
    {
        public int Id { get; set; }
        public string? NotifyId { get; set; }
        public string? Reference { get; set; }
        public string? Uri { get; set; }
        public string? TemplateId { get; set; }
        public string? TemplateUri { get; set; }
        public int? TemplateVersion { get; set; }
        public string? EmailResponseContentFrom { get; set; }
        public string? EmailResponseContentBody { get; set; }
        public string? EmailResponseContentSubject { get; set; }

        public string? ExceptionCode { get; set; }
        public string? ExceptionMessage { get; set; }

        public int EmailLogId { get; set; }
        public EmailLog EmailLog { get; set; } = null!;
    }</code></pre><p></p><p>EmailTriggerActivation - this is used to hold emails that are triggered for processing once a delivered email has been confirmed.  So this allows us to activate the TP enquiry emails when the enquirer email has been delivered.</p><pre><code>    public class EmailTriggerActivation
    {
        public int Id { get; set; }
        public int EmailLogId { get; set; }
        public int ActivateEmailLogId { get; set; }

        public EmailLog EmailLog { get; set; } = null!;
        public EmailLog ActivateEmailLog { get; set; } = null!;
    }</code></pre><p></p><p>EmailStatus - this is a table that holds the status of the email (e.g. is it ready for processing, was there an immediate error sending it, was it not delivered, was it delivered etc).  See full list of email statuses below.</p><pre><code>    public class EmailStatus
    {
        public int Id { get; set; }

        public string Status { get; set; } = null!;

        public string Description { get; set; } = null!;

        public bool AllowEmailSending { get; set; } //If the email should be sent (in conjunction with EmailLog.ProcessFromDate and EmailLog.FinishProcessingDate being valid)

        public bool PollForStatusUpdateIfSent { get; set; } //If we should check Notify for status updates (in conjunction with EmailLog.LastEmailSendAttemptDate and EmailLog.FinishProcessingDate being valid)

        public int? RetrySendInSeconds { get; set; } //If the sending should be retired against this status, how many seconds until retry (updates the EmailLog.ProcessFromDate)

        public ICollection&lt;EmailLog&gt;? EmailLogs { get; set; }
    }</code></pre><p></p><p>The following FK’s will also be added to the existing Domain models, so we know what each email relates to:</p><pre><code>Enquiry.EnquirerEnquirySubmittedEmailLogId
TuitionPartnerEnquiry.TuitionPartnerEnquirySubmittedEmailLogId
EnquiryResponse.TuitionPartnerResponseEmailLogId
EnquiryResponse.EnquirerResponseEmailLogId</code></pre><p></p><p>Database ERD of the above classes:</p><div data-node-type="mediaSingle" data-layout="align-start" data-width=""><div data-id="cfaf0282-e89a-4efc-8572-5678c53bc0ed" data-node-type="media" data-type="file" data-collection="upload-user-collection--1471422447" data-width="1140" data-height="861" data-alt="" title="Attachment" style="display: inline-block; border-radius: 3px; background: #EBECF0; box-shadow: 0 1px 1px rgba(9, 30, 66, 0.2), 0 0 1px 0 rgba(9, 30, 66, 0.24);" data-file-name="image-20230419-095642.png" data-file-size="140339" data-file-mime-type="image/png" data-context-id="57285"></div></div><p></p><p>Suggested Email Statuses:</p><p>Statuses 5-10 below are from <a data-inline-card="" href="https://docs.notifications.service.gov.uk/net.html#email-status-descriptions" data-card-data="">https://docs.notifications.service.gov.uk/net.html#email-status-descriptions</a> </p>

Id | Status | Description | AllowEmailSending | PollForStatusUpdateIfSent | RetrySendInSeconds
-- | -- | -- | -- | -- | --
1 | to-be-processed | Has been newly added to log, will be processed next time the email processing is run | True | True | NULL
2 | waiting-to-be-triggered | Is waiting for a chained email to be delivered (e.g. TP emails are only sent once the enquirer email has been delivered) | True | True | NULL
3 | delayed-email | The email is to be sent is to be sent in the future (e.g. send notification emails daily) | True | True | NULL
4 | been-processed | The email has been processed and sent to GOV.UK Notify to be delivered | False | True | NULL
5 | created | GOV.UK Notify status: has placed the message in a queue, ready to be sent to the provider. It should only remain in this state for a few seconds. | False | True | NULL
6 | sending | GOV.UK Notify status: has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. GOV.UK Notify is waiting for delivery information. | False | True | NULL
7 | delivered | GOV.UK Notify status: the message was successfully delivered. | False | False | NULL
8 | permanent-failure | GOV.UK Notify status: the provider could not deliver the message because the email address was wrong. You should remove these email addresses from your database. | False | False | NULL
9 | temporary-failure | GOV.UK Notify status: the provider could not deliver the message. This can happen when the recipientâ€™s inbox is full or their anti-spam filter rejects your email. Check your content does not look like spam before you try to send the message again. | True | True | 600
10 | technical-failure | GOV.UK Notify status: your message was not sent because there was a problem between Notify and the provider. You’ll have to try sending your messages again. | True | True | 60
11 | processing-failure | Error when calling the GOV.UK Notify SendEmailAsync() | True | False | 600
12 | sending-emails-deactivated | The emails were configured not to be sent at the time the email was processed, expected to be used for non-production environments only | False | False | NULL

<p></p><p>Following implementation the following performance was tested:</p><ul class="ak-ul"><li><p>The polling of the emails to get updated status is approximately 10 emails = 1 second.</p></li><li><p>The sending of emails is approximately 10 emails = 1 second.</p></li><li><p>This means the polling would take approximately 5 seconds (e.g. 50 TP responses) to send all the emails for a single enquiry to the TPs.  With the subsequent polling taking 5 seconds to update the delivery statuses.</p></li><li><p>This means that so long as there are 5 or fewer enquiries per minute the polling should be able to handle them if we poll for updates every minute, without hitting issues with the subsequent polling starting.</p></li></ul>
</body>
</html>

## Guidance to review

A thorough code review and test are needed.

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1151

## Things to check

- [ ] **PR deployment has been signed off by wider team


UI/Pages/Admin/ProcessEmails.cshtml.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        [BindProperty] public ProcessedEmailsModel Data { get; set; } = new();
+
+        public async Task OnGetAsync()
+        {
+            var request = new ProcessEmailsCommand()
+            {
+                //Errored email = "763404e0-47ed-4b72-93a7-843d31597206"
+                //OK Email = "708fec24-0a92-4089-8e6c-9b10f68a8526"
+                NotificationId = "763404e0-47ed-4b72-93a7-843d31597206"
```
hasan3d, line 23: We don't need the `NotificationId` here. Can we remove that?

Application/Commands/Admin/ProcessEmailsCommand.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Application.Common.Interfaces;
+using Application.Common.Models.Admin;
+
+namespace Application.Commands.Admin;
+
+public record ProcessEmailsCommand : IRequest<ProcessedEmailsModel>
+{
+    public string? NotificationId { get; set; } = null!;
```
hasan3d, line 8: Do we need this?

UI/Pages/Admin/ProcessEmails.cshtml.cs
kieranohanlon: Good shout, old testing code that has no impact but certain should have been removed.  Done.

Application/Commands/Admin/ProcessEmailsCommand.cs
kieranohanlon: Good shout, old testing code that has no impact but certain should have been removed.  Done.

Infrastructure/Configuration/FeatureFlags.cs
```diff
@@ -6,5 +6,8 @@ public class FeatureFlags
 
     public bool VerifyEmail { get; set; } = true;
     public bool EnquiryBuilder { get; set; } = true;
+    public bool SendEmailsFromNtp { get; set; } = true;
+    public bool VerifyEmail { get; set; } = true;
```
hasan3d, line 10: We don't need to add `VerifyEmail` as we already have it.

vars-staging.yml
```diff
@@ -12,11 +12,14 @@ govuk-notify-enquiry-submitted-to-tp-template-id: b24ffca7-6119-4dbf-8ae2-dceb2c
...
 govuk-notify-email-verification-template-id: badaacff-87b8-4046-b953-a85a0a9e2a8b
-override-email-address: 
-email-amalgamate-responses: false
+email-override-email-address: 
+email-merge-responses: false
 email-all-sent-to-enquirer: true
 feature-flag-verify-email: true
 feature-flag-enquiry-builder: true
+feature-flag-send-emails-from-ntp: true
+feature-flag-verify-email: true
```
hasan3d, line 21: Can we remove this? As we already have it.

vars-research.yml
```diff
@@ -12,11 +12,14 @@ govuk-notify-enquiry-submitted-to-tp-template-id: b24ffca7-6119-4dbf-8ae2-dceb2c
...
 govuk-notify-email-verification-template-id: badaacff-87b8-4046-b953-a85a0a9e2a8b
-override-email-address: 
-email-amalgamate-responses: true
+email-override-email-address: 
+email-merge-responses: true
 email-all-sent-to-enquirer: true
 feature-flag-verify-email: true
 feature-flag-enquiry-builder: true
+feature-flag-send-emails-from-ntp: true
+feature-flag-verify-email: true
```
hasan3d, line 21: Can we remove this? As we already have it.

vars-qa.yml
```diff
@@ -12,11 +12,14 @@ govuk-notify-enquiry-submitted-to-tp-template-id: b24ffca7-6119-4dbf-8ae2-dceb2c
...
 govuk-notify-email-verification-template-id: badaacff-87b8-4046-b953-a85a0a9e2a8b
-override-email-address: 
-email-amalgamate-responses: true
+email-override-email-address: 
+email-merge-responses: true
 email-all-sent-to-enquirer: true
 feature-flag-verify-email: true
 feature-flag-enquiry-builder: true
+feature-flag-send-emails-from-ntp: true
+feature-flag-verify-email: true
```
hasan3d, line 21: Can we remove this? As we already have it.

vars-production.yml
```diff
@@ -12,11 +12,14 @@ govuk-notify-enquiry-submitted-to-tp-template-id: b24ffca7-6119-4dbf-8ae2-dceb2c
...
 govuk-notify-email-verification-template-id: badaacff-87b8-4046-b953-a85a0a9e2a8b
-override-email-address: 
-email-amalgamate-responses: false
+email-override-email-address: 
+email-merge-responses: false
 email-all-sent-to-enquirer: false
 feature-flag-verify-email: true
 feature-flag-enquiry-builder: true
+feature-flag-send-emails-from-ntp: true
+feature-flag-verify-email: true
```
hasan3d, line 21: Can we remove this? As we already have it.

vars-pr.yml
```diff
@@ -12,11 +12,14 @@ govuk-notify-enquiry-submitted-to-tp-template-id: b24ffca7-6119-4dbf-8ae2-dceb2c
...
 govuk-notify-email-verification-template-id: badaacff-87b8-4046-b953-a85a0a9e2a8b
-override-email-address: 
-email-amalgamate-responses: true
+email-override-email-address: 
+email-merge-responses: true
 email-all-sent-to-enquirer: true
 feature-flag-verify-email: true
 feature-flag-enquiry-builder: true
+feature-flag-send-emails-from-ntp: true
+feature-flag-verify-email: true
```
hasan3d, line 21: Can we remove this? As we already have it.

Infrastructure/Configuration/FeatureFlags.cs
kieranohanlon: Good spot - sorted

vars-staging.yml
kieranohanlon: Good spot - sorted

vars-research.yml
kieranohanlon: Good spot - sorted

hasan3d (approved):
lgtm!

Infrastructure/Services/ProcessEmailsService.cs
```diff
@@ -0,0 +1,486 @@
...
+                        //Save per record rather than all in one go, so that it will process other emails and not fail for all
+                        await _unitOfWork.Complete();
+
+                        processedEmailsModel.EmailsUpdatedStatus++;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    //We don't want to throw the error here, we log and continue processing other emails
+                    _logger.LogError(ex, "Unexpected error in emailsToPollForStatusUpdate for EmailLog.ClientReferenceNumber: {ClientReferenceNumber}, EmailLog.Id: {EmailLogId}", emailLog.ClientReferenceNumber, emailLog.Id);
```
hasan3d, line 280: I am seeing the following error on my IDE: `Dereference of a possibly null reference` for the `emailLog.ClientReferenceNumber`. I suggest using the following `emailLog!.ClientReferenceNumber`.

kieranohanlon: Will add this in to NTP-1218 which is in dev at the moment, since this has been fully tested.

hasan3d (approved):
lgtm

PR #356, opened 2023-06-14
NTP-1197: Reduce responses from TPs that do not meet the schools needs - where the TPs are responding "no"

kieranohanlon:
## Context

Reduce responses from TPs that do not meet the schools needs - where the TPs are responding "no"

We are introducing a read-only screen of the schools enquiries which TPs will view before they view the existing page where they actually craft and send their response. When the TPs open the schools enquiries, this new page (or new view) is the they will visit first from their e-mail notification about the enquiry.

On this page, they will have CTA buttons to select their intention; replying or not replying, with the negative having a short ‘are you sure’ journey, and the positive taking the TP to the page/view they use at the moment to make the response.

When the schools view their manage enquiry screen, they will be able to see how many TPs have chosen not to respond. To allow for this change, The section which tells the school how many TPs the enquiry was sent to will be updated more generally.

## Guidance to review

As a School

Given I have sent an enquiry

When I view the replies I have from TPs

Then They are all valuable - they are all there because they can meet my requirement, or have reasonable substitutes. They don’t say “no” to my non-negotiables such as SEND support, or perhaps timing before an exam period.

## Link to Jira ticket

https://dfedigital.atlassian.net.mcas.ms/browse/NTP-1197

## Things to check


Domain/TuitionPartnerEnquiry.cs
```diff
@@ -9,6 +9,10 @@ public class TuitionPartnerEnquiry
 
     public int MagicLinkId { get; set; }
 
+    public bool TuitionPartnerDecinedEnquiry { get; set; } = false;
```
hasan3d, line 12: Typo here. It should be `TuitionPartnerDeclinedEnquiry`

Domain/TuitionPartnerEnquiry.cs
```diff
@@ -9,6 +9,10 @@ public class TuitionPartnerEnquiry
...
+
+    public DateTime? TuitionPartnerDecinedEnquiryDate { get; set; }
```
hasan3d, line 14: Typo here. It should be `TuitionPartnerDeclinedEnquiryDate `

Domain/TuitionPartnerEnquiry.cs
kieranohanlon: Good shout - have updated all instances of "Decined", including the migration script name, which will break this PR when tries to run it, so will create new PR to run the db migrations from scratch.

Domain/TuitionPartnerEnquiry.cs
kieranohanlon: Good shout - have updated all instances of "Decined", including the migration script name, which will break this PR when tries to run it, so will create new PR to run the db migrations from scratch.

PR #357, opened 2023-06-15
NOJIRA: Adding the ADR for using Azure BLOB Storage

ashward:
## Context

Adding the ADR for the decision to use BLOB storage


hasan3d (approved):
lgtm!

PR #362, opened 2023-06-22
NTP-1197: Reduce responses from TPs that do not meet the schools needs - where the TPs are responding "no"

kieranohanlon:
## Context

Reduce responses from TPs that do not meet the schools needs - where the TPs are responding "no"

We are introducing a read-only screen of the schools enquiries which TPs will view before they view the existing page where they actually craft and send their response. When the TPs open the schools enquiries, this new page (or new view) is the they will visit first from their e-mail notification about the enquiry.

On this page, they will have CTA buttons to select their intention; replying or not replying, with the negative having a short ‘are you sure’ journey, and the positive taking the TP to the page/view they use at the moment to make the response.

When the schools view their manage enquiry screen, they will be able to see how many TPs have chosen not to respond. To allow for this change, The section which tells the school how many TPs the enquiry was sent to will be updated more generally.

## Guidance to review

As a School

Given I have sent an enquiry

When I view the replies I have from TPs

Then They are all valuable - they are all there because they can meet my requirement, or have reasonable substitutes. They don’t say “no” to my non-negotiables such as SEND support, or perhaps timing before an exam period.

## Link to Jira ticket

https://dfedigital.atlassian.net.mcas.ms/browse/NTP-1197

## Things to check

- [x] Database migrations can be applied to the current codebase in main without causing exceptions
- [x] Debug logging has been added after all logic decision points


Application/Commands/Enquiry/Respond/AddEnquiryResponseCommand.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System;
```
hasan3d, line 2: We don't this as we already have it.

Application/Commands/Enquiry/Respond/DeclineEnquiryResponseCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+    public DeclineEnquiryResponseCommandHandler(IUnitOfWork unitOfWork,
+        ILogger<DeclineEnquiryResponseCommandHandler> logger)
+    {
+        _unitOfWork = unitOfWork;
+        _logger = logger;
+    }
+
+    public async Task<bool> Handle(DeclineEnquiryResponseCommand request, CancellationToken cancellationToken)
+    {
+        var result = new ResponseConfirmationModel();
```
hasan3d, line 27: Can we remove the unused local variable `result`?

Application/Commands/Enquiry/Respond/DeclineEnquiryResponseCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var errorMessage = $"Previously declined TuitionPartnerEnquiry with Support Ref ('{request.SupportReferenceNumber}') and Tuition Partner SeoUrl ('{request.TuitionPartnerSeoUrl}')";
+            _logger.LogError(errorMessage);
+            throw new ArgumentException(errorMessage);
+        }
+
+        tpEnquiry.TuitionPartnerDeclinedEnquiry = true;
+        tpEnquiry.TuitionPartnerDeclinedEnquiryDate = DateTime.UtcNow;
+
+        await _unitOfWork.Complete();
```
hasan3d, line 59: I suggest we add a try catch here and catch any exception and log and throw appropriate error.

Application/Commands/Enquiry/Respond/AddEnquiryResponseCommand.cs
kieranohanlon: Done

Application/Commands/Enquiry/Respond/DeclineEnquiryResponseCommand.cs
kieranohanlon: Done

Application/Commands/Enquiry/Respond/DeclineEnquiryResponseCommand.cs
kieranohanlon: I don't think that we should be catching and logging unless there is a good reason for doing so - e.g. you're having to handle the exception in some way or logging additional data.  Since when it propagates up to the UI it will get logged anyway.

But this did highlight that in a few of the other command handlers we were doing that, without adding any specific logging data or handling the exception.  So I have removed the unnecessary error handling from those, so it's cleaner and the handlers are aligned.  If you look at the commit and let me know if you have issues with it.

Tests/Tests.csproj
```diff
@@ -35,6 +35,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\GiasPostcodeSearch\GiasPostcodeSearch.csproj" />
```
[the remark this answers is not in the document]
kieranohanlon: Good call, not sure how or why I added this, removed.

hasan3d (approved):
lgtm

PR #365, opened 2023-06-22
NTP-1244: Fix URL creation if a TP has forward slash in their name and tweak alerts

kieranohanlon:
## Context

Fix URL creation if a TP has forward slash in their name and tweak alerts

A new TP has the name “Ravenspoint Ltd T/A Tutor Doctor Reading, Maidenhead, Wokingham“ the forward slash is causing an issue since we use this name to create the URL for the TP details.  So the / should be stripped out when used in the URL, otherwise the path is invalid.

We also spotted that if there is a missing logo we are not alerted (this is logged as info).  Since all TPs seem to have a logo we will log this as a warning (so will not stop upload, but we’ll be alerted), so we have visibility of the missing logo now that Tribal control the data maintenance for TPs.

Since the new spreadsheet successfully uploaded, we had an issue rolling back to the previous version.  This is because the old data could not be uploaded and was throwing an error, in case this is being done by accident.  To resolve this, we will set this as a warning, allowing old data to be uploaded, with a longer term solution being added to the backlog: https://dfedigital.atlassian.net/browse/NTP-1245

## Guidance to review

Run against new TP data with / and also make sure can rollback

## Link to Jira ticket

https://dfedigital.atlassian.net/browse/NTP-1244

## Things to check

### DIFF
--- a/Application/Extensions/StringExtensions.cs
+++ b/Application/Extensions/StringExtensions.cs
@@ -11,9 +11,17 @@ namespace Application.Extensions;
 public static class StringExtensions
 {
     [return: NotNullIfNotNull("value")]
-    public static string? ToSeoUrl(this string? value)
+    public static string? ToSeoUrl(this string? value, bool replaceForwardSlash = true)
     {
-        var seo = value?
+        if (string.IsNullOrEmpty(value))
+            return value;
+
+        var seo = value;
+
+        if (replaceForwardSlash)
+            seo = seo.Replace("/", "");
+
+        seo = seo
             .RegexReplace(StringConstants.CamelCaseBoundaries, " $1")
             .Trim()
             .RegexReplace(StringConstants.SpacesAndUnderscore, "-")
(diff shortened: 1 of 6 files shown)


Application/Extensions/StringExtensions.cs
```diff
@@ -11,9 +11,17 @@ namespace Application.Extensions;
...
+    public static string? ToSeoUrl(this string? value, bool replaceForwardSlash = true)
     {
-        var seo = value?
+        if (string.IsNullOrEmpty(value))
+            return value;
+
+        var seo = value;
+
+        if (replaceForwardSlash)
+            seo = seo.Replace("/", "");
```
hasan3d, line 22: Are we expecting forward slash for the Seo Url?

hasan3d: If not we can incorporate the forward slash replacing logic with the code below.

kieranohanlon: I had originally put it within the `UrlUnsafeCharacters` - but on certain occasions we do need to leave the forward slashes, if part of full path.  See `ToSeoUrl_ReturnsKebabCase_WhenDirectory` and `SeoRouteConvention` within solution,

hasan3d: I see what you mean it's also used with directory, make sense.

hasan3d (approved):
lgtm

PR #399, opened 2023-08-21
Revert "NTP-1292: Fix Inconsistent Domain Resolution for https://www.find-tuition-partner.service.gov.uk"

hasan3d:
Reverts DFE-Digital/find-a-tuition-partner#398


hasan3d:
Reverting the changes as we created the DNS zone and associated Front Door custom domain without the Terraform on the production.

